PR #9, opened 2019-12-13
pocket: add duration time for test

cwen0:
Signed-off-by: cwen0 <cwenyin0@gmail.com> 

add duration time to run pocket，so we can set a finish time for this test

### DIFF
--- a/pocket/core/core.go
+++ b/pocket/core/core.go
@@ -1,22 +1,24 @@
 package core
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"path"
 	"regexp"
 	"sync"
-	"path"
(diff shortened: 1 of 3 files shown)


pocket/core/core.go
```diff
@@ -177,7 +179,11 @@ func (e *Executor) Start() error {
 			go e.startBinlogTestDataCompare()
 		}
 	}
-	return nil
+
+	select {
+	case <-ctx.Done():
+		return nil
```
you06: It is not considered block here, return here will not stop the executation. May start a go routine to wait for ctx.Done signal and call `Stop` function.

PR #14, opened 2020-01-07
Improvement of SQL generator

you06:

- generator interface
- support online DDL mode
- configuration check duration
- configuration init table count
- default value for int/float/varchar/timestamp/datetime
- add a data compare tool which can find difference by uuid
- support general log in config


go-sqlsmith/dml_sql.go
```diff
@@ -12,25 +12,26 @@ import (
...
+// InsertStmt implement insert statement from AST
+func (s *SQLSmith) InsertStmt(fn bool) (string, string, error) {
 	tree := s.insertStmt()
 	return s.Walk(tree)
 }
 
-// InsertStmt make random insert statement SQL
-func (s *SQLSmith) InsertStmt(fn bool) (string, error) {
+// InsertStmtStr make random insert statement SQL
+func (s *SQLSmith) InsertStmtStr(fn bool) (string, string, error) {
```
mahjonp: Seems there is no function will call InsertStmtStr?

go-sqlsmith/stateflow/schema.go
```diff
@@ -59,22 +59,34 @@ func (s *StateFlow) randTableFromTable(table *types.Table, newName bool, fn bool
...
-		if index == k {
-			for len(newTable.Columns) == 0 {
-				newTable = s.randTableFromTable(table, newName, fn)
-			}
-			return newTable
+func (s *StateFlow) randTable(newName bool, fn bool, online bool) (*types.Table) {
+	tables := []*types.Table{}
+	for _, table := range s.db.Tables {
+		// get online tables
+		if !online || !s.db.Online || table.Online {
```
mahjonp: I'm confused, db.Online and table.Online wasn't always set to same boolean value? 😶

go-sqlsmith/types/table.go
```diff
@@ -7,12 +7,16 @@ import (
...
-	Type string
-	Columns map[string]*Column
-	Indexes []string
+	Type        string
+	Columns     map[string]*Column
+	Indexes     []string
+	// Online is for self obtain
+	Online      bool
+	// OnlineOther is for other instances obtain
+	OnlineOther bool
```
mahjonp: what are the meaning of `Online` and `OnlineOther`? 😆

you06:
@mahjonp I'll add some comments later, and this PR is not ready in this time.

go-sqlsmith/dml_sql.go
you06: Yes, but there are 2 ways of generate insert statement, from AST and building from string directly. Just keep both of them here.

zhouqiang-cl:
Can we review this pr now @you06

go-sqlsmith/data_gen_test.go
```diff
@@ -16,13 +16,13 @@ package sqlsmith
 import "testing"
 
 func TestSQLSmith_GenData(t *testing.T) {
```
mapleFU: Why this is leaved but all commented?

you06:
@zhouqiang-cl Still some code tidy up works to do, there will be some updates tomorrow.

go-sqlsmith/data_gen_test.go
you06: There should be, please review later.

go-sqlsmith/ddl_ast.go
```diff
@@ -26,9 +26,9 @@ func (s *SQLSmith) createTableStmt() ast.Node {
 		Constraints: []*ast.Constraint{},
 		Options: []*ast.TableOption{},
 	}
-	if util.Rd(4) == 0 {
-		createTableNode.Partition = s.partitionStmt()
-	}
+	// if util.Rd(4) == 0 {
+	// 	createTableNode.Partition = s.partitionStmt()
```
mahjonp: go-sqlsmith should add more switches in future.

go-sqlsmith/stateflow/schema.go
you06: They are not, this is for avoiding online DDL, when start a txn, db.Online and some tables will be set to true, and the other tables has Online value false, then the go-sqlsmith should generate DDL SQLs only about these tables during this txn.

go-sqlsmith/types/table.go
you06: `Online` means this table is online and will be manipulated in a txn, `OnlineOther` means this table is manipulated in other transactions which should not be a DDL table.

PR #27, opened 2020-02-05
*: extract interface to be implemented

mahjonp:

- [x] Extract cluster `provisioner` interface as the adapter of cluster manager on K8s.
- ~~[ ] Refactor the `Nemesis` interface~~

To retain the backward compatibility, I extract out a `Node` struct and denotes a Pod on K8s environment.

That's means we cannot inject pod-kill kind faults using chaos-mesh currently, because they destroy the Pod and make the related `Node` be illegal. we need resolve this limitation in the future.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/cmd/rawkv/main.go
+++ b/cmd/rawkv/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"flag"
+	"github.com/pingcap/tipocket/pkg/cluster"
 	"log"
 	"time"
 
@@ -54,5 +55,5 @@ func main() {
 		Nemesises:     *nemesises,
 		VerifySuit:    verifySuit,
 	}
-	suit.Run(context.Background(), []string{})
+	suit.Run(context.Background(), []cluster.Node{})
 }
--- a/cmd/rawkv/main.go
+++ b/cmd/rawkv/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"flag"
-	"github.com/pingcap/tipocket/pkg/cluster"
 	"log"
 	"time"
 
@@ -55,5 +54,5 @@ func main() {
 		Nemesises:     *nemesises,
 		VerifySuit:    verifySuit,
 	}
-	suit.Run(context.Background(), []cluster.Node{})
+	suit.Run(context.Background())
 }
--- a/cmd/tidb/main.go
+++ b/cmd/tidb/main.go
@@ -46,12 +46,12 @@ func main() {
 	switch *clientCase {
 	case "bank":
 		creator = tidb.BankClientCreator{}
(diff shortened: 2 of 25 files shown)


cmd/rawkv/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"flag"
+	"github.com/pingcap/tipocket/pkg/cluster"
```
mapleFU: group imports

cmd/tidb/main.go
```diff
@@ -46,12 +46,12 @@ func main() {
 	switch *clientCase {
 	case "bank":
```
mapleFU: So, in m1 we only need to support `bank` and `long_fork`?

cmd/util/suit.go
```diff
@@ -2,7 +2,7 @@ package util
 
 import (
 	"context"
-	"fmt"
+	"github.com/pingcap/tipocket/pkg/cluster"
```
mapleFU: ditto

cmd/util/suit.go
```diff
@@ -18,6 +18,8 @@ import (
 // Suit is a basic chaos testing suit with configurations to run chaos.
 type Suit struct {
 	*control.Config
+	// Provisioner deploy the SUT cluster
```
mapleFU: What do you mean by "SUT"?

db/cluster/cluster.go
```diff
@@ -3,6 +3,7 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"github.com/pingcap/tipocket/pkg/cluster"
```
mapleFU: ditto

pkg/cluster/cluster.go
```diff
@@ -0,0 +1,25 @@
...
+	PodName string
+
+	IP   string
+	Port string
+}
+
+// Provisioner provides a collection of APIs to deploy/destroy a cluster
+type Provisioner interface {
+	// SetUp sets up cluster, returns err or all nodes info
+	SetUp(ctx context.Context, spec json.RawMessage) (error, []Node)
```
you06: Will this interface change in the future, compare with RawMessage input, I prefer a struct or an interface.

mahjonp: yep, you're right. I'll change it.

cmd/util/suit.go
mahjonp: SUT means `system under test`.

cmd/tidb/main.go
mahjonp: yeah.

PR #31, opened 2020-02-09
Replace nemesis to k8s style

mapleFU:
Signed-off-by: mapleFU <1506118561@qq.com>

Replace nemesis to k8s style.

```
KUBECONFIG=~/.kube/config ./bin/chaos-tidb -namespace fxw -hub hub.pingcap.net -nemesis random_kill
```

Change nemesis code to golang k8s style.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/control/control.go
```diff
@@ -158,22 +158,22 @@ func (c *Controller) syncExec(f func(i int)) {
 func (c *Controller) setUpDB() {
 	log.Printf("begin to set up database")
 	c.syncExec(func(i int) {
```
you06: `syncExec` will trigger input function for every pod, includes tikv, pd, tidb, we should set up db for tidb pod only here. A further problem is that all tidb pods are under a same nodeport service which runs load balance, we should not set up db for the same db address multiple times.

pkg/core/nemesis.go
```diff
@@ -16,9 +23,9 @@ type Nemesis interface {
 	// TearDown(ctx context.Context, node string) error
 
 	// Invoke executes the nemesis
-	Invoke(ctx context.Context, node cluster.Node, args ...string) error
+	Invoke(ctx context.Context, node cluster.Node, chaosNS string, args ...string) error
```
mahjonp: Adding `chaoNS` isn't necessary.

pkg/control/control.go
```diff
@@ -139,7 +138,7 @@ ROUND:
 	nemesisWg.Wait()
 
 	c.tearDownClient()
-	c.tearDownDB()
+	//c.tearDownDB()
```
mahjonp: revert?

pkg/control/control.go
```diff
@@ -78,7 +78,6 @@ func (c *Controller) Close() {
 
 // Run runs the controller.
 func (c *Controller) Run() {
-	// c.setUpDB()
```
mahjonp: uncomment?

pkg/nemesis/nemesis.go
```diff
@@ -2,34 +2,49 @@ package nemesis
...
+		return err
+	}
+	log.Printf("Creating pod-kill with node %s(ns:%s)\n", node.PodName, node.Namespace)
+	return podChaos(ctx, c, node.Namespace, node.Namespace, node.PodName, v1alpha1.PodFailureAction)
 }
 
 func (kill) Recover(ctx context.Context, node cluster.Node, args ...string) error {
-	db := core.GetDB(args[0])
-	return db.Start(ctx, node)
+	c, err := createClient()
```
mahjonp: inject Chaos client into kill struct?

mapleFU: emm, good idea.

cmd/tidb/main.go
```diff
@@ -29,6 +29,7 @@ var (
 	checkerNames = flag.String("checker", "porcupine", "checker name, eg, porcupine, tidb_bank_tso")
 	pprofAddr    = flag.String("pprof", "0.0.0.0:8080", "Pprof address")
 	namespace    = flag.String("namespace", "tidb-cluster", "test namespace")
+	//chaosNamespace = flag.String("chaos-ns", "chaos-testing", "test chaos namespace")
```
you06: Why not set a flag for chaos namespace?

pkg/core/nemesis.go
```diff
@@ -105,7 +115,8 @@ func (NoopNemesisGenerator) Generate(nodes []cluster.Node) []*NemesisOperation {
 	ops := make([]*NemesisOperation, len(nodes))
```
you06: Does this means the amount of nemesis is equal to amount of nodes?

mapleFU: @you06 Sure, you can take a look at the code below.

mahjonp: Yes, NoopNemesisGenerator is a faker generator, `NoopNemesis` does nothing actually.

PR #33, opened 2020-02-10
*: implement k8s deployment

you06:

Implement k8s deployment

```
make tidb
KUBECONFIG=~/.kube/config ./bin/chaos-tidb -namespace test-tidb -image-version v3.0.9
```

### Check List 

Tests 

 - Manual test

Code changes

 - Has Go code change

Side effects

Related changes


you06:
@mahjonp @mapleFU there are not any comments in the original test-infra, so I do not comment inside it also, PTAL

cmd/util/suit.go
```diff
@@ -53,10 +56,12 @@ func (suit *Suit) Run(ctx context.Context) {
 
 	sctx, cancel := context.WithCancel(ctx)
 
-	err, suit.Config.Nodes = suit.Provisioner.SetUp(sctx, nil)
+	suit.Config.Nodes, err = suit.Provisioner.SetUp(sctx, suit.Cluster)
 	if err != nil {
 		log.Fatalf("deploy a cluster failed, err: %s", err)
 	}
+	log.Println("deploy cluster success")
+	os.Exit(0)
```
mapleFU: Whats this?

mahjonp: remove debug stubs?

pkg/test-infra/pkg/tidb/recommend.go
```diff
@@ -58,7 +61,20 @@ func (t *TiDBClusterRecommendation) TiDBReplicas(replicas int32) *TiDBClusterRec
...
+	}
+	b.WriteString(fixture.E2eContext.DockerRepository)
+	b.WriteString("/")
+	b.WriteString(name)
+	b.WriteString(":")
+	b.WriteString(fixture.E2eContext.ImageVersion)
+	return b.String()
+}
+
+// RecommendedTiDBCluster does a recommendation, tidb-operator do not have same defaults yet
```
mapleFU: Does this publish the svc for tidb cluster?

pkg/test-infra/pkg/tidb/recommend.go
```diff
@@ -58,7 +61,20 @@ func (t *TiDBClusterRecommendation) TiDBReplicas(replicas int32) *TiDBClusterRec
 	return t
 }
 
-// tidb-operator do not have sane defaults yet, do a recommendation
+func buildImage(name string) string {
```
mapleFU: `buildImage`  is ambiguity...

mapleFU:
Why just have a constexpr set for strings, like tidb/mysql client name

you06:
> Why just have a constexpr set for strings, like tidb/mysql client name

Do you mean this code?

```golang
Ports: []corev1.ServicePort{
	{
		Name: "mysql-client",
		Port: 4000,
	},
	{
		Name: "status",
		Port: 10080,
	},
}
```

pkg/test-infra/pkg/tidb/recommend.go
you06: `*corev1.Service` field in `TiDBClusterRecommendation` will create a nodeport which will public svc for tidb cluster

pkg/test-infra/pkg/tidb/recommend.go
you06: Do you have some advice for the function name?

PR #34, opened 2020-02-11
cases: make bank workload runnable with k8s

you06:

### Check List 

Tests 

 - Manual test

Code changes

 - Has Go code change

```release-note
NONE
```


pkg/control/control.go
```diff
@@ -78,7 +78,7 @@ func (c *Controller) Close() {
 
 // Run runs the controller.
 func (c *Controller) Run() {
-	c.setUpDB()
+	// c.setUpDB()
```
mahjonp: we can use NoopDB here.

PR #35, opened 2020-02-13
Control client concurrency

mahjonp:

On K8s, test accesses TiDB server through nodePort service, there was one clientNode(tidb service), and so only one client concurrency.

The pr adds the clientCount option to control the client concurrencies.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

```
NONE
```

Related changes

```
NONE
```

```release-note
NONE
```


db/tidb/bank.go
```diff
@@ -33,8 +33,9 @@ type bankClient struct {
 	accountNum int
 }
 
-func (c *bankClient) SetUp(ctx context.Context, nodes []cluster.ClientNode, node cluster.ClientNode) error {
+func (c *bankClient) SetUp(ctx context.Context, nodes []cluster.ClientNode, idx int) error {
 	c.r = rand.New(rand.NewSource(time.Now().UnixNano()))
+	node := nodes[idx]
 	db, err := sql.Open("mysql", fmt.Sprintf("root@tcp(%s:%d)/test", node.IP, node.Port))
 	if err != nil {
 		return err
```
you06: Seems `c.db` will be overwrite here

pkg/control/control.go
```diff
@@ -320,7 +318,7 @@ func (c *Controller) onNemesisLoop(ctx context.Context, index int, op *core.Neme
 	case <-time.After(op.RunTime):
 	case <-ctx.Done():
 	}
-	if err := nemesis.Recover(ctx, node, op.RecoverArgs...); err != nil {
+	if err := nemesis.Recover(context.TODO(), node, op.RecoverArgs...); err != nil {
```
mahjonp: I use the context.TODO() here to ensure nemesis.Recover wouldn't be interrupted by the global timeout context, but it's better to use another Timeout context.

mahjonp: Emm, maybe retry is better. timeout context isn't suitable.

PR #36, opened 2020-02-13
Network partition: add partition_one

mahjonp:

partition_one isolates one node from other nodes

### Check List 

Tests 

Code changes

 - Has Go code change

```release-note
NONE
```


cmd/util/suit.go
```diff
@@ -9,11 +9,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/pingcap/tipocket/pkg/nemesis"
```
mapleFU: Why do you move it here...

pkg/nemesis/network_partition.go
```diff
@@ -0,0 +1,160 @@
...
+					Pods: map[string][]string{
+						anotherPart[0].Namespace: extractPodNames(anotherPart),
+					},
+				},
+				TargetMode: chaosv1alpha1.AllPodMode,
+			},
+		},
+	})
+}
+
```
mapleFU: The `NetworkChaos` templates is almost the same...why not just create a function like `podTag`

mahjonp: IMO, it's early to extract such function

cmd/util/suit.go
mahjonp: fixed

pkg/nemesis/network_partition.go
```diff
@@ -0,0 +1,165 @@
+package nemesis
+
+import (
+	"context"
+	"fmt"
+	uuid "github.com/satori/go.uuid"
```
mapleFU: group imports

you06:
I have the question about kill nemesis, it kills randomly from all nodes, including tidb node. Does the bank case work when killing tidb?

mahjonp:
> I have the question about kill nemesis, it kills randomly from all nodes, including tidb node. Does the bank case work when killing tidb?

Yeah, I think BankClient can solve such connection error.

pkg/nemesis/network_partition.go
mahjonp: resolved

PR #37, opened 2020-02-13
.*: support binlog and pocket workload

you06:

Support binlog in k8s and pocket workload

Add a control mode which does not rely on the default scheduler which invoke every round.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


you06:
This PR is ready but should merge #40 and #41 first, because the interface change will make the above two PRs not work. I should make some changes after #40 and #41 got merged.

zhouqiang-cl:
@you06 please solve conflict

you06:
@mahjonp @mapleFU PTAL

cmd/tidb/main.go
```diff
@@ -102,6 +105,10 @@ func main() {
 		log.Fatalf("invalid checker %s", *checkerNames)
 	}
 
+	if withProf {
+		cfg.Mode = control.ModeWithPerf
+	}
+
```
you06: Set `Mode` field to `ModeWithPerf` to enable perf.

cmd/util/suit.go
```diff
@@ -79,6 +79,7 @@ func (suit *Suit) Run(ctx context.Context) {
 	if len(suit.Config.ClientNodes) == 0 {
 		log.Panic("no client nodes exist")
 	}
+	suit.Config.ClientCount = len(suit.Config.ClientNodes)
```
mahjonp: Why need this?

db/tidb/bank.go
```diff
@@ -203,12 +203,16 @@ func (c *bankClient) DumpState(ctx context.Context) (interface{}, error) {
 	return balances, nil
 }
 
+func (c *bankClient) Start(ctx context.Context, cfg interface{}, clientNodes []clusterTypes.ClientNode) error {
+	return nil
```
mahjonp:
```suggestion
	panic("unreachable")
```

pkg/test-infra/pkg/tidb/operation.go
```diff
@@ -447,3 +522,41 @@ func getDiscoveryMeta(tc *v1alpha1.TidbCluster) (metav1.ObjectMeta, label.Label)
...
+
+func (t *TidbOps) parseNodeFromPodList(pods *corev1.PodList) []clusterTypes.Node {
+	var nodes []clusterTypes.Node
+	for _, pod := range pods.Items {
+		if strings.Contains(pod.ObjectMeta.Name, "discovery") {
+			continue
+		}
+		nodes = append(nodes, clusterTypes.Node{
+			Namespace: pod.ObjectMeta.Namespace,
+			PodName:   pod.ObjectMeta.Name,
```
mahjonp: Need fill the `Component` field here,

https://github.com/pingcap/tipocket/blob/6dba244c5f1eed302f5e2928463eb90ad1051d09/pkg/cluster/k8s.go#L115-L129

pkg/test-infra/pkg/tidb/recommend.go
```diff
@@ -18,15 +18,18 @@ import (
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type TiDBClusterRecommendation struct {
 	*v1alpha1.TidbCluster
 	*corev1.Service
+	NS   string
+	Name string
```
mahjonp: NS and Name are same as TidbCluster.ObjectMeta.Namespace and TidbCluster.ObjectMeta.Name, so I think we can remove these two new fields. What's your opinion?

cmd/util/suit.go
you06: Maybe we should remove this field

pkg/test-infra/pkg/tidb/recommend.go
you06: In binlog cluster, two tidb cluster with different names are in the same namespace.

https://github.com/pingcap/tipocket/blob/005f3f1824075707165fbc32dcf1da671c2f4f79/pkg/test-infra/pkg/binlog/recommander.go#L49-L50

PR #41, opened 2020-02-18
support service quality

mahjonp:

Measure failure requests count before and after nemesis was injected. 
When the avg failure count of one continuous minute is below 0.8 * the failure count before nemesis, we say the the system recover to p80 line from that time point.  And p90 or p99 are similar.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cmd/util/suit.go
```diff
@@ -25,7 +25,8 @@ type Suit struct {
 	core.ClientCreator
 	// nemesis, seperated by comma.
 	Nemesises string
-
+	// perform service quality checking
+	WithProf   bool
```
you06: Is `WithProf` a common config field for all the cases?

you06: I add a field to control the schedule mode as `control.Config.Mode`, maybe we can use it later by add a `ModeWithProf` value.

db/tidb/bank.go
```diff
@@ -68,6 +68,13 @@ func (c *bankClient) SetUp(ctx context.Context, nodes []cluster.ClientNode, idx
 }
 
 func (c *bankClient) TearDown(ctx context.Context, nodes []cluster.ClientNode, idx int) error {
+	if idx != 0 {
+		return nil
+	}
+	sql := `drop table if exists accounts`
+	if _, err := c.db.Exec(sql); err != nil {
+		return err
+	}
```
you06: Does it really necessary? We start a new cluster for every experiment.

cmd/util/suit.go
mahjonp: I wish it could be a common config field for all cases.

> I add a field to control the schedule mode as `control.Config.Mode`, maybe we can use it later by add a `ModeWithProf` value.

Good idea.

db/tidb/bank.go
mahjonp: It's useful when I test using a pre-deployed cluster.

pkg/test-infra/pkg/control/control.go
```diff
@@ -166,7 +166,7 @@ func (c *Controller) dumpState(ctx context.Context, recorder *history.Recorder)
 	defer cancel()
 
 	for _, client := range c.clients {
-		for _, node := range c.cfg.ClientNodes {
+		for _, node := range c.cfg.Nodes {
```
mahjonp: we will remove control in test-infra in the future.

pkg/nemesis/pod_kill.go
```diff
@@ -38,8 +38,9 @@ func podKillNodes(nodes []cluster.Node, n int) []*core.NemesisOperation {
 	for i := 0; i < n; i++ {
 		ops[indices[i]] = &core.NemesisOperation{
 			Type:        core.PodKill,
-			InvokeArgs:  []interface{}{nodes[i], freq},
-			RecoverArgs: []interface{}{nodes[i]},
+			Node:        &nodes[i],
```
mahjonp: For kill, and netem chaos, I record target node here, this change aims to simplify the nemesis apply phase.

pkg/nemesis/pod_kill.go
```diff
@@ -53,45 +54,43 @@ func NewPodKillGenerator(name string) core.NemesisGenerator {
...
 
+// podKill implements Nemesis
 type podKill struct {
 	k8sNemesisClient
 }
 
-func (k podKill) Invoke(ctx context.Context, _ cluster.Node, args ...interface{}) error {
-	node, freq := extractPodKillArgs(args)
+func (k podKill) Invoke(ctx context.Context, node *cluster.Node, args ...interface{}) error {
+	freq := extractPodKillArgs(args)
```
you06:
```suggestion
	freq := extractPodKillArgs(args...)
```

pkg/test-infra/pkg/tidb/operation.go
```diff
@@ -297,6 +299,36 @@ func (t *TidbOps) ApplyDiscovery(tc *v1alpha1.TidbCluster) error {
 	return nil
 }
 
+func (t *TidbOps) GetPDMember(namespace, name string) (string, []string, error) {
```
mahjonp: TiDB-cluster track and record the PD leader and member at TC.status.PD.Members, so I use that information to retrieve the leader and members currently.

pkg/nemesis/pod_kill.go
```diff
@@ -53,45 +54,43 @@ func NewPodKillGenerator(name string) core.NemesisGenerator {
...
 	podChaos := podKillTag(freq, node.Namespace, node.Namespace,
-		node.PodName, chaosv1alpha1.PodKillAction)
+		node.PodName)
 	return k.cli.ApplyPodChaos(ctx, &podChaos)
 }
 
-func (k podKill) Recover(ctx context.Context, _ cluster.Node, args ...interface{}) error {
-	node := extractKillArgs(args)
+func (k podKill) Recover(ctx context.Context, node *cluster.Node, args ...interface{}) error {
+	freq := extractPodKillArgs(args)
```
you06:
```suggestion
	freq := extractPodKillArgs(args...)
```

pkg/nemesis/pod_kill.go
```diff
@@ -53,45 +54,43 @@ func NewPodKillGenerator(name string) core.NemesisGenerator {
...
```
mahjonp:
```suggestion
	freq := extractPodKillArgs(args...)
```

db/tidb/bank_service_quality.go
```diff
@@ -0,0 +1,209 @@
...
+	var startTime time.Time
+	var endTime time.Time
+	for idx := range ops {
+		op := ops[idx]
+		if op.Action == core.ReturnOperation {
+			response, ok := op.Data.(bankResponse)
+			if !ok {
+				continue
+			}
+			if response.Unknown {
```
you06: Where does unknown type response come from? Why not drop it?

db/tidb/bank_service_quality.go
```diff
@@ -0,0 +1,209 @@
...
+				if !op.Time.IsZero() {
+					series = append(series, op.Time)
+				}
+			}
+			if startTime.IsZero() {
+				startTime = op.Time
+			}
+			if endTime.IsZero() {
+				endTime = op.Time
+			}
```
you06: Redundant code.

mahjonp: resolved

db/tidb/bank_service_quality.go
mahjonp: Unknown type response usually means timeout or failure, I treat them as failure response here.

mahjonp:
PTAL again @you06

db/tidb/bank_service_quality.go
```diff
@@ -0,0 +1,206 @@
...
+			break
+		} else if op.Action == core.RecoverNemesis {
+			//recoveryTime = op.Time
+			//break
+		}
+		idx++
+	}
+	failCountPerMinuteBeforeInject = failCount / injectionTime.Sub(startTime).Minutes()
+	if idx == len(ops) {
+		log.Panic("expect more events after recovery from nemesis")
```
mapleFU: I think just return is ok.

db/tidb/bank_service_quality.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	var window *countWindow
+	metrics := rtoMetrics{nemesis: &nemesisRecord}
+
+	window, metrics.totalMeasureTime = convertToFailSeries(ops[idx:])
+	if window != nil {
+		for true {
+			startTime, failCount, err := window.count(duration)
```
mapleFU: making it an iterable object maybe more understandable.

db/tidb/bank_service_quality.go
```diff
@@ -0,0 +1,206 @@
...
+
+			if 0.8*failCountPerMinute <= failCountPerMinuteBeforeInject && p80 == nil {
+				p80 = &dur
+			}
+			if 0.9*failCountPerMinute <= failCountPerMinuteBeforeInject && p90 == nil {
+				p90 = &dur
+			}
+			if 0.99*failCountPerMinute <= failCountPerMinuteBeforeInject && p99 == nil {
+				p99 = &dur
+				break
```
mapleFU: So, p99 is better than p90, and p90 is better than p80. What abort worse case?

pkg/control/control.go
```diff
@@ -142,7 +144,74 @@ ROUND:
...
+			go func() {
+				defer nemesisWg.Done()
+				time.Sleep(time.Second * 10)
+				c.dispatchNemesisWithRecord(ctx, g, recorder)
+			}()
+
+			clientWg.Wait()
+			log.Printf("%s round %d client requests done", g.Name(), round)
+			cancel()
+			nemesisWg.Wait()
```
mapleFU: Why didn't nemesis done before `clinetWg`?

cmd/tidb/main.go
```diff
@@ -79,10 +80,14 @@ func main() {
 
 	parser := tidb.BankParser()
 	model := tidb.BankModel()
+	withProf := false
```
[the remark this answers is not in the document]
mahjonp: yeah.

db/tidb/bank.go
```diff
@@ -328,9 +335,9 @@ func (p bankParser) OnRequest(data json.RawMessage) (interface{}, error) {
 func (p bankParser) OnResponse(data json.RawMessage) (interface{}, error) {
 	r := bankResponse{}
 	err := json.Unmarshal(data, &r)
-	if r.Unknown {
-		return nil, err
-	}
+	//if r.Unknown {
```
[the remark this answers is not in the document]
mahjonp: We need the response time recorded on it. In the original implementation, Unknown type response records was discarded here and rebuilt again at https://github.com/pingcap/tipocket/blob/06ac82ebbfab63b9f79e5d737a5bffbc5d1cc952/pkg/history/history.go#L198, but that may cause the lost of response time when re-create that will cause we can't get occur time of each failure response.

pkg/control/control.go
mahjonp: The sequence doesn't matter here. I just ensure that verify is after clients and nemesis all finished.

db/tidb/bank_service_quality.go
mahjonp: That's a good idea

PR #45, opened 2020-02-24
Add container kill support

mahjonp:

Add container kill support, that can be used on short-term node killing case.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/db/tidb/bank.go
+++ b/db/tidb/bank.go
@@ -93,7 +93,7 @@ func (c *bankClient) invokeRead(ctx context.Context, r bankRequest) bankResponse
 
 	rows, err := txn.QueryContext(ctx, "select balance from accounts")
 	if err != nil {
-		return bankResponse{Unknown: true}
+		return bankResponse{Unknown: true, Error: err.Error()}
 	}
 	defer rows.Close()
 
@@ -101,7 +101,7 @@ func (c *bankClient) invokeRead(ctx context.Context, r bankRequest) bankResponse
 	for rows.Next() {
 		var v int64
 		if err = rows.Scan(&v); err != nil {
-			return bankResponse{Unknown: true}
+			return bankResponse{Unknown: true, Error: err.Error()}
 		}
 		balances = append(balances, v)
 	}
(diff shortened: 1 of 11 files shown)


pkg/nemesis/container_kill.go
```diff
@@ -0,0 +1,113 @@
...
+			Namespace: chaosNs,
+		},
+		Spec: chaosv1alpha1.PodChaosSpec{
+			Selector: chaosv1alpha1.SelectorSpec{
+				Pods: pods,
+			},
+			ContainerName: containerName,
+			Action:        chaos,
+			Mode:          chaosv1alpha1.OnePodMode,
+			Scheduler:     &chaosv1alpha1.SchedulerSpec{Cron: "@every 1000000s"}, // let it be scheduled in our control
```
mapleFU: Seems this is regard a instant action?

db/tidb/bank.go
```diff
@@ -153,7 +153,7 @@ func (c *bankClient) Invoke(ctx context.Context, node cluster.ClientNode, r inte
 	}
 
 	if err = txn.Commit(); err != nil {
-		return bankResponse{Unknown: true, Tso: tso, FromBalance: fromBalance, ToBalance: toBalance}
+		return bankResponse{Unknown: true, Tso: tso, FromBalance: fromBalance, ToBalance: toBalance, Error: err.Error()}
```
mapleFU: does Error means unexpected error like io error?

pkg/nemesis/container_kill.go
mahjonp: Sure, I think that using freq to control schedule such instant action is better.

db/tidb/bank.go
mahjonp: Yeah, mostly timeout error because of some fault injections, eg:

* Error 9005: Region is unavailable
* Error 9002: TiKV server timeout

mahjonp:
> ContainerKill and PodFailure LGTM. But I'm confused on what `convertToFailSeries` changed

I misunderstand Unknown type response, the original process was correct, so I revert it.

https://github.com/pingcap/tipocket/blob/ae941e7980049a27570582e18dedb1da43864e6c/db/tidb/bank.go#L340-L342

After reverting, the program processes Unknown type by moving all of this kind of records on the tail of history from original position, so I use the occur time of previous InvokeOperation record of these Unknown type response as the failure time.

PR #51, opened 2020-02-26
QoS: use TPCC TPM as QoS metric

mahjonp:

Uses TPCC success OPM as the QoS metric. I add the `ticker` property to control request frequency of each client to achieve the stable OPM, and define that OPM value as the baseline. 

After fault injected, the duration of recovering to 80% of baseline is recorded as p80, p90 and p99 are similar.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


db/tidb/bank_qos.go
```diff
@@ -13,15 +13,36 @@ import (
...
 	p90              time.Duration
 	p99              time.Duration
-	totalMeasureTime time.Duration
 }
 
-type bankServiceQualityChecker struct {
-	profFile string
+func (m *rtoMetrics) Record(filePath string) error {
+	os.MkdirAll(path.Dir(""), 0755)
+	f, err := os.OpenFile(filePath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
```
mapleFU: does it need to be `O_SYNC`?

pkg/core/checker.go
```diff
@@ -22,3 +22,30 @@ func (NoopChecker) Check(m Model, ops []Operation) (bool, error) {
 func (NoopChecker) Name() string {
 	return "NoopChecker"
 }
+
+type multiChecker struct {
+	checkers []Checker
```
mapleFU: Maybe a `build(...args)` method or append is better...

mahjonp: `MultiChecker` is the build method

db/tidb/bank_qos.go
mahjonp: It's more safe if so but we don't need that guarantee here.

pkg/core/checker.go
mapleFU: Got

cmd/tidb/main.go
```diff
@@ -40,12 +40,12 @@ var (
 	runTime      = flag.Duration("run-time", 100*time.Minute, "client test run time")
 	clientCase   = flag.String("case", "bank", "client test case, like bank,multi_bank")
 	historyFile  = flag.String("history", "./history.log", "history file")
+	qosFile      = flag.String("qosFile", "./qos.log", "qos file")
```
mapleFU: using qos-file to unify style?

cmd/tpcc/main.go
```diff
@@ -17,10 +17,13 @@ import (
 	"context"
 	"flag"
 	"log"
+	"math"
 	"net/http"
 	_ "net/http/pprof"
 	"time"
 
+	"github.com/pingcap/tipocket/pkg/core"
```
mapleFU: Group it with others.

cmd/tpcc/main.go
```diff
@@ -32,11 +35,14 @@ import (
 
 var (
 	clientCount  = flag.Int("client", 5, "client count")
-	requestCount = flag.Int("request-count", 1000, "client test request count")
+	requestCount = flag.Int("request-count", math.MaxInt64, "client test request count")
```
mapleFU: Why?

cmd/tpcc/main.go
```diff
@@ -32,11 +35,14 @@ import (
...
 	round        = flag.Int("round", 3, "client test request round")
-	runTime      = flag.Duration("run-time", 100*time.Minute, "client test run time")
+	ticker       = flag.Duration("ticker", time.Second, "ticker control request emitting freq")
+	runTime      = flag.Duration("run-time", 10*time.Minute, "client test run time")
 	historyFile  = flag.String("history", "./history.log", "history file")
+	qosFile      = flag.String("qosFile", "./qos.log", "qos file")
```
mapleFU: Same, qos-file

cmd/tpcc/main.go
```diff
@@ -67,25 +74,40 @@ func main() {
...
 	}
-
 	clientCreator := &tidb.TPCCClientCreator{}
 	creator := clientCreator
+	var checker core.Checker
+	switch *checkerName {
+	case "consistency":
+		checker = &tidb.TPCCChecker{CreatorRef: clientCreator}
+	case "qos":
+		checker = core.MultiChecker("TPCC&Qos", tidb.TPCCQosChecker(time.Minute, *qosFile), &tidb.TPCCChecker{CreatorRef: clientCreator})
```
mapleFU: "TPCC&Qos"?

cmd/util/suit.go
```diff
@@ -118,7 +97,162 @@ func (suit *Suit) Run(ctx context.Context) {
...
+		node clusterTypes.ClientNode,
+		proc *int64,
+		requestCount *int64,
+		recorder *history.Recorder) {
+		log.Printf("begin to run command on node %s", node)
+
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		token := make(chan interface{}, 1)
```
mapleFU: `chan struct{}` maybe better.

mapleFU: Why 1?

cmd/tpcc/main.go
mahjonp: In this case, I use `runTime` to control the total run time of each round, so I adjust the default value to a very big number to prevent each round finishes too quickly.

pkg/core/checker.go
```diff
@@ -22,3 +22,30 @@ func (NoopChecker) Check(m Model, ops []Operation) (bool, error) {
...
+	name     string
+}
+
+func (c multiChecker) Check(m Model, ops []Operation) (valid bool, err error) {
+	for _, checker := range c.checkers {
```
mapleFU: Does it's better to run concurrency?

mahjonp: It's better but not necessary. That's not a RT job. :)

PR #77, opened 2020-03-11
*: use test fixture

mahjonp:

There are much of boilerplate code when create a new case, so ugly. 

I use the test fixture to gather common flags to simplify the difficulty when add a new test case.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

None

Related changes

```release-note
NONE
```


Makefile
```diff
@@ -54,9 +54,10 @@ on-dup:
 fmt: groupimports
 	go fmt ./...
 
-mod:
+tidy:
+	@echo "go mod tidy"
 	GO111MODULE=on go mod tidy
-	@git diff --exit-code -- go.sum go.mod
+	@git diff --exit-code -- go.mod
```
mahjonp: I'm not sure the reason it cannot pass in travis CI. I delete go.sum and run `make tidy` to generate go.sum but hasn't any difference between previous one.

pkg/test-infra/tests/client.go
```diff
@@ -23,42 +24,35 @@ import (
...
 	CDC    *cdc.CdcOps
 	MySQL  *mysql.MySQLOps
 	TiDB   *tidb.TidbOps
 	Binlog *binlog.Ops
 }
 
-// NewE2eCli creates e2e client
-func NewE2eCli(conf *rest.Config) *E2eCli {
+// NewTestCli creates test client
+func NewTestCli(conf *rest.Config) *TestCli {
```
mapleFU: Why do you remove Chaos and BR supports?

pkg/util/ssh/ssh.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	verbose = flag.Bool("ssh-verbose", false, "show the verbose of SSH command")
+	verbose bool
```
mapleFU: Why?

cmd/verifier/main.go
```diff
@@ -14,13 +25,12 @@ import (
...
 	"github.com/pingcap/tipocket/pkg/model"
+	"github.com/pingcap/tipocket/pkg/test-infra/fixture"
 	"github.com/pingcap/tipocket/pkg/verify"
 )
 
 var (
-	historyFile = flag.String("history", "./history.log", "history file")
-	names       = flag.String("names", "", "model names, seperate by comma")
-	pprofAddr   = flag.String("pprof", "0.0.0.0:6060", "Pprof address")
+	names = flag.String("names", "", "model names, seperate by comma")
```
mapleFU: `seperated`

pkg/test-infra/tests/client.go
mahjonp: Chaos was useless, BR we haven't support now.

pkg/util/ssh/ssh.go
mahjonp: because it will register a flag to every binary, you can see it from --help. Most of our cases run on K8s, so `ssh-verbose` is redundant now.

mapleFU:
```
func init() {
	flag.BoolVar(&verbose, "ssh-verbose", false, "show the verbose of SSH command")
}
```

When package is import, init will be executed.

mahjonp: Yes, so only on need.

PR #83, opened 2020-03-16
support loki query

yeya24:

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/loki/loki.go
```diff
@@ -0,0 +1,83 @@
...
+	query := fmt.Sprintf(`{container_name="%s", namespace="%s"} %s"%s"`,
+		containerName, ns, op, match)
+
+	var nonEqual string
+	for _, v := range nonMatch {
+		nonEqual += fmt.Sprintf(` !="%s"`, v)
+	}
+	query += nonEqual
+
+	res, err := c.cli.QueryRange(query, 1000, time.Now().Truncate(lookbackTime), time.Now(), logproto.BACKWARD, 15*time.Second, true)
```
mahjonp: is `Add` more suitable than `Truncate` here?

yeya24: You mean `Add(-lookbackTime)`? What's the difference between the two?

mahjonp: Truncate returns the result of rounding t down to a multiple of d (since the zero time)...

yeya24: You are right. Will update soon.

pkg/loki/loki.go
```diff
@@ -0,0 +1,89 @@
...
+
+// NewLokiClient creates a client to query loki.
+func NewLokiClient(address, username, password string) *lokiClient {
+	return &lokiClient{
+		cli: &client.Client{
+			Address:  address,
+			Username: username,
+			Password: password,
+		},
+		startTime: time.Now(),
```
you06: Seems the client will only fetch logs after `startTime`, if there are some logs being collected into Loki before the client init, we can't access them.

yeya24: Yes. So the loki client needs to be initialized first.

pkg/loki/loki.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	query += nonEqual
+
+	queryTo := time.Now()
+	queryFrom := queryTo.Add(-lookbackTime)
+	if queryFrom.Before(c.startTime) {
+		queryFrom = c.startTime
+	}
+
+	res, err := c.cli.QueryRange(query, 1000, queryFrom, queryTo, logproto.BACKWARD, 15*time.Second, true)
```
you06: What will be if there are more than 1000 results in this duration?

pkg/loki/loki.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, v := range nonMatch {
+		nonEqual += fmt.Sprintf(` !="%s"`, v)
+	}
+	query += nonEqual
+
+	queryTo := time.Now()
+	queryFrom := queryTo.Add(-lookbackTime)
+	if queryFrom.Before(c.startTime) {
+		queryFrom = c.startTime
+	}
```
you06: Just an advice, if `lookbackTime` is set to 0, it will query all logs in the history, unless it query between the given duration.

pkg/loki/loki.go
you06: If we want get all logs from loki, why not just set the start time to 0?

pkg/loki/loki.go
yeya24: If there are more than 1000 line of logs, it will return 1000. Making this parameter more than 1000 is not recommended. If allow more too many lines of logs, then it will cause a huge stress to loki itself.

you06: Should we handle this situation and make multi requests for each batch?

pkg/loki/loki.go
yeya24: No. I want to avoid the case that there already some history logs in loki with the same namespace. So just query from the time that the loki client initialized.

yeya24: But maybe we should use a better way to do this. Maybe add another label to do filtering.

you06: Get it, we can keep this solution by now.

pkg/loki/loki.go
yeya24: What do you mean by `query all logs`? But thanks, I checked the code in loki and I really need to handle this case. If `starttime == endtime`, it will return an error.

pkg/loki/loki.go
yeya24: I don't think we need. But maybe @mahjonp has some other ideas. IMO we just use logs to grep some key words like `error` or `panic`. Btw, the `lookbackTime` should not be too long, I don't think it can exceed `1d`, then it will too slow.

pkg/loki/loki.go
you06: I was considering setting `startTime` to 0 for all logs, but here we should avoid to get logs from previous cases. Maybe we can just handle the case `queryFrom == queryTo` and return empty result for now.

mahjonp: Or maybe we can change the method signature? pass startTime and endTime maybe more convenient?

pkg/loki/loki.go
you06: this LGTM

mahjonp: So the 1000 here is the filtering log limit or the scan log lines limit? if it's the first one, I think it's big enough

mahjonp: > I don't think we need. But maybe @mahjonp has some other ideas. IMO we just use logs to grep some key words like `error` or `panic`. Btw, the `lookbackTime` should not be too long, I don't think it can exceed `1d`, then it will too slow.

Yes, it's usually small. Our case is running round by round, and each round isn't too long in normal.

pkg/loki/loki.go
yeya24: Pass startTime and endTime is ok. But the startTime check is required unless we can find a way to just query the cluster we want.

pkg/loki/loki.go
yeya24: The limit of the results. Not the logs loki scans.

PR #85, opened 2020-03-17
k8s cluster: Support tidb monitor

mahjonp:

Need to create a TiDB monitor when create the TiDB cluster.

On GKE, I found a problem that when I use LoadBalancer as the Grafana service type, the NodePort is always changing, and the Grafana dashboard isn't accessible. Maybe refs: https://github.com/pingcap/tidb-operator/issues/1949

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/test-infra/tidb/recommend.go
+++ b/pkg/test-infra/tidb/recommend.go
@@ -175,7 +175,7 @@ func RecommendedTiDBCluster(ns, name string) *TiDBClusterRecommendation {
 						Version:   "v1.0.1",
 					},
 				},
-				ImagePullPolicy: corev1.PullAlways,
+				ImagePullPolicy: corev1.PullIfNotPresent,
 			},
 		},
 		Service: &corev1.Service{
--- a/pkg/test-infra/tidb/recommend.go
+++ b/pkg/test-infra/tidb/recommend.go
@@ -26,7 +26,8 @@ import (
 )
 
 type TiDBClusterRecommendation struct {
-	*v1alpha1.TidbCluster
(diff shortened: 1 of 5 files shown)


pkg/test-infra/tidb/recommend.go
```diff
@@ -88,7 +89,7 @@ func RecommendedTiDBCluster(ns, name string) *TiDBClusterRecommendation {
 				Namespace: ns,
 				Labels: map[string]string{
 					"app":      "tipocket-tidbcluster",
-					"instance": "name",
```
yeya24: Good catch! 😄

pkg/test-infra/tidb/recommend.go
```diff
@@ -129,6 +130,54 @@ func RecommendedTiDBCluster(ns, name string) *TiDBClusterRecommendation {
...
+						Version:   "v3.0.5",
+					},
+				},
+				Reloader: v1alpha1.ReloaderSpec{
+					MonitorContainer: v1alpha1.MonitorContainer{
+						BaseImage: "pingcap/tidb-monitor-reloader",
+						Version:   "v1.0.1",
+					},
+				},
+				ImagePullPolicy: corev1.PullAlways,
```
yeya24: Well. Is it better to use `IfnotPresent`?

mahjonp: You are right, `IfNotPresent` is better, I will update it.

PR #87, opened 2020-03-18
Add panic logs check

yeya24:

Use a goroutine to check panic logs from loki.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cmd/util/suit.go
```diff
@@ -90,6 +92,8 @@ func (suit *Suit) Run(ctx context.Context) {
 		suit.NemesisGens,
 		suit.ClientRequestGen,
 		suit.VerifySuit,
+		loki.NewLokiClient(fixture.Context.LokiAddress,
```
mahjonp: How about use a interface here?

pkg/control/control.go
```diff
@@ -54,6 +57,7 @@ func NewController(
 	nemesisGenerators []core.NemesisGenerator,
 	clientRequestGenerator func(ctx context.Context, client core.Client, node clusterTypes.ClientNode, proc *int64, requestCount *int64, recorder *history.Recorder),
 	verifySuit verify.Suit,
+	lokiCli *loki.LokiClient,
```
mahjonp: How about use an interface here?

pkg/test-infra/fixture/fixture.go
```diff
@@ -16,9 +16,8 @@ package fixture
 import (
 	"flag"
 	"net/http"
-	"time"
-
 	_ "net/http/pprof" // pprof
+	"time"
```
mapleFU: Just put them together is OK

pkg/control/control.go
yeya24: I think it is ok to use lokiClient for now. But let me add a todo. What do you think

PR #96, opened 2020-03-19
Support ABTest cluster provider and a ABTest case

you06:
This PR enable abtest and add k8s provider for double cluster.

### DIFF
--- a/pkg/test-infra/abtest/operation.go
+++ b/pkg/test-infra/abtest/operation.go
@@ -1,0 +1,113 @@
+// Copyright 2019 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package abtest
+
+import (
+	"sync"
+
+	"github.com/pingcap/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	clusterTypes "github.com/pingcap/tipocket/pkg/cluster/types"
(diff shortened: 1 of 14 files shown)


you06:
In my test, it leads error very quickly due to the data inconsistency. The log looks weird and I still need some time to find out why it failed.

pkg/test-infra/abtest/operation.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Apply abtest cluster
+func (t *Ops) Apply(tc *Recommendation) error {
+	var (
+		wg   sync.WaitGroup
+		errs []error
+	)
+	wg.Add(2)
+
+	go func() {
```
mahjonp: use [errgroup.Group{}](https://godoc.org/golang.org/x/sync/errgroup) here?

pkg/test-infra/fixture/fixture.go
```diff
@@ -165,6 +166,9 @@ func init() {
 	flag.StringVar(&Context.LokiAddress, "loki-addr", "", "loki address. If empty then don't query logs from loki.")
 	flag.StringVar(&Context.LokiUsername, "loki-username", "", "loki username. Needed when basic auth is configured in loki")
 	flag.StringVar(&Context.LokiPassword, "loki-password", "", "loki password. Needed when basic auth is configured in loki")
+
+	flag.StringVar(&Context.ABTestConfig.Cluster1Version, "cluster1-version", "", "specify version for cluster1")
```
mahjonp:
```suggestion
	flag.StringVar(&Context.ABTestConfig.Cluster1Version, "abtest.a.image-name", "", "specify version for cluster1")
```

mahjonp:
Rest LGTM

pkg/test-infra/fixture/fixture.go
you06: I think "abtest.a.version" would be better, cause it only defines the cluster version but not the whole image name.

mahjonp: OK for me

you06:
@mahjonp @mapleFU PTAL again

PR #101, opened 2020-03-21
*: redefine fixture.Context.Purge to reuse exist clusters

mahjonp:

We use fixture.Context.PurgeOnSuccess to control delete namespace before, but found that that it's not enough because we also need another flag to control whether or not to destroy current cluster, eg. when run multiple cases on one cluster.

So in this PR, I refine the `Purge` flag and use it to reach the above two objective at the same time.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/test-infra/abtest/operation.go
+++ b/pkg/test-infra/abtest/operation.go
@@ -19,14 +19,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	clusterTypes "github.com/pingcap/tipocket/pkg/cluster/types"
-	"github.com/pingcap/tipocket/pkg/test-infra/fixture"
 	"github.com/pingcap/tipocket/pkg/test-infra/tidb"
 )
 
 // Ops knows how to operate TiDB with binlog on k8s
 type Ops struct {
-	cli        client.Client
-	tidbClient *tidb.TidbOps
+	cli client.Client
+	*tidb.TidbOps
 }
 
 // New creates binlog ops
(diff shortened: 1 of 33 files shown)


pkg/test-infra/abtest/operation.go
```diff
@@ -37,48 +37,45 @@ func New(cli client.Client, tidbClient *tidb.TidbOps) *Ops {
...
 	})
 	g.Go(func() error {
-		return t.ApplyTiDBCluster(tc.Cluster2, tidb.ConfigMaps{
-			Tikv: fixture.Context.ABTestConfig.TiKV2ConfigMap,
-			Tidb: fixture.Context.ABTestConfig.TiDB2ConfigMap,
-			Pd:   fixture.Context.ABTestConfig.PD2ConfigMap,
+		return t.ApplyTiDBCluster(tc.Cluster2, tidb.Config{
+			Tikv: fixture.Context.ABTestConfig.TiKVConfigFile,
+			Tidb: fixture.Context.ABTestConfig.TiDBConfigFile,
+			Pd:   fixture.Context.ABTestConfig.PDConfigFile,
```
you06: If b cluster's config is empty, should we use the same config file as a cluster?

mahjonp: it's also reasonable for me that use the default config to start the b cluster when the b cluster's config is empty.

you06: LGTM

PR #102, opened 2020-03-21
Rename config flag

yeya24:

This pr renames several configs related flags. Since they are just the path of the config file, so it is more appropriate to rename them to `xxx-config` instead of `xxx-configmap`

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/test-infra/fixture/fixture.go
```diff
@@ -170,12 +170,12 @@ func init() {
...
-	flag.StringVar(&Context.TiDBConfigFile, "tidb-configmap", "", "path of tidb configmap (cluster A in abtest case)")
-	flag.StringVar(&Context.TiKVConfigFile, "tikv-configmap", "", "path of tikv configmap (cluster A in abtest case)")
-	flag.StringVar(&Context.PDConfigFile, "pd-configmap", "", "path of pd configmap (cluster A in abtest case)")
-	flag.StringVar(&Context.ABTestConfig.TiDBConfigFile, "abtest.b.tidb-configmap", "", "tidb configmap for cluster b")
-	flag.StringVar(&Context.ABTestConfig.TiKVConfigFile, "abtest.b.tikv-configmap", "", "tikv configmap for cluster b")
-	flag.StringVar(&Context.ABTestConfig.PDConfigFile, "abtest.b.pd-configmap", "", "pd configmap for cluster b")
+	flag.StringVar(&Context.TiDBConfigFile, "tidb-config", "", "path of tidb config file (cluster A in abtest case)")
+	flag.StringVar(&Context.TiKVConfigFile, "tikv-config", "", "path of tikv config file (cluster A in abtest case)")
+	flag.StringVar(&Context.PDConfigFile, "pd-config", "", "path of pd config file (cluster A in abtest case)")
+	flag.StringVar(&Context.ABTestConfig.TiDBConfigFile, "abtest.b.tidb-config", "", "tidb config file for cluster b")
```
mahjonp: And how about rename `abtest.b.xxx` as `abtest.xxx`? eg, we use tidb-config as `abtest.a.tidb-config` in fact, so I think we can remove the `b.` prefix here?

pkg/test-infra/fixture/fixture.go
```diff
@@ -170,12 +170,12 @@ func init() {
...
+	flag.StringVar(&Context.ABTestConfig.TiKVConfigFile, "abtest.b.tikv-config", "", "tikv config file for cluster b")
+	flag.StringVar(&Context.ABTestConfig.PDConfigFile, "abtest.b.pd-config", "", "pd config file for cluster b")
 	flag.StringVar(&Context.ABTestConfig.Cluster1Version, "abtest.a.version", "", "specify version for cluster a")
```
mahjonp: If you think is ok, I think we can remove `abtest.a.version` and rename `abtest.b.version` to `abtest.image_version` too.

pkg/test-infra/fixture/fixture.go
yeya24: Yes, sounds reasonable.

pkg/test-infra/fixture/fixture.go
yeya24: Done

PR #107, opened 2020-03-22
abtest: fix inconsistency

you06:

ABTest will fail by inconsistency, I guess it is due to setting global variables required some seconds to take effect. This PR fix this problem by waiting 10 seconds after setting up during database client init.

### DIFF
--- a/pkg/pocket/pkg/logger/logger.go
+++ b/pkg/pocket/pkg/logger/logger.go
@@ -56,7 +56,7 @@ func (l *Logger) writeLine(line string) error {
 	if l.mute {
 		return nil
 	} else if l.print {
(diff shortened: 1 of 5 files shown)


pkg/pocket/pkg/logger/logger.go
```diff
@@ -56,7 +58,7 @@ func (l *Logger) writeLine(line string) error {
 	if l.mute {
 		return nil
 	} else if l.print {
-		log.Info(line)
+		log.Info(fmt.Sprintf("[%s] %s", l.name, line))
```
you06: When the log is printed to the stdout, I use the name filed to specify which client the log is belong to.

PR #112, opened 2020-03-22
Add resolve lock test for green gc

MyonKeminta:
Added a test case to check GC Resolve Lock's correctness under Green GC.

I think this PR is not being merged immediately since it uses a specific version of TiDB.


tests/resolve-lock/resolve_lock.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+
+	for i := 0; i < loadDataConcurrency; i++ {
+		err = <-errCh
+		if err != nil {
+			return errors.Trace(err)
+		}
+	}
+
+	// TODO: Is this the correct way to create the PD client and kv client?
```
mahjonp: `nodes` only contain tidb pod, so you will get an empty set here now.

mahjonp: maybe we need inject all nodes here. 😲

MyonKeminta: Is there any way to access PD and TiKV?

Makefile
```diff
@@ -19,7 +19,7 @@ all: build
 
 chaos: tidb
 
-build: fmt tidb pocket tpcc ledger txn-rand-pessimistic on-dup sqllogic block-writer \
+build: fmt tidb pocket tpcc ledger resolve_lock txn-rand-pessimistic on-dup sqllogic block-writer \
```
Illyrix: do you mean resolve-lock?

MyonKeminta: Sorry.

tests/resolve-lock/resolve_lock.go
```diff
@@ -0,0 +1,493 @@
...
+			}()
+		}
+
+		key = nextScanKeyCopy
+		if len(key) == 0 {
+			break
+		}
+	}
+
+	for currentRunningRegions >= c.GenerateLockConcurrency {
```
youjiali1995, line 272: It's useless. Do you intend to leave some goroutines running during resolving locks?

tests/resolve-lock/resolve_lock.go
```diff
@@ -0,0 +1,493 @@
...
+				return errors.Trace(err)
+			}
+			currentRunningRegions--
+		}
+	}
+
+	return nil
+}
+
+func (c *resolveLockClient) singleLockRegion(ctx context.Context, startKey, endKey []byte) error {
```
youjiali1995, line 287: I think only one lock is not enough...

mahjonp:
close PR because it's duplicated with https://github.com/pingcap/tipocket/pull/202

PR #118, opened 2020-03-23
Fix: a/b test test config dir copy problem

mahjonp:

In previous PR, I create a subdir on configmaps, but current Dockerfile won't copy that.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,11 @@ WORKDIR /src
 COPY go.mod .
 COPY go.sum .
 
+RUN go mod download
+
 COPY . .
+
+RUN make build
 
 FROM alpine:3.8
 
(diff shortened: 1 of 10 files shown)


Dockerfile
```diff
@@ -8,19 +8,15 @@ WORKDIR /src
 COPY go.mod .
 COPY go.sum .
 
-RUN go mod download
-
 COPY . .
 
-RUN make build
```
Illyrix: `/src/bin/` would be empty without `make build`?

mahjonp: Sorry, I forget to recover them after debugging.

PR #126, opened 2020-03-24
Add tikv read error

fewdan:

add readerr_tikv

add readerr_tikv

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


pkg/test-infra/tidb/recommend.go
```diff
@@ -107,7 +110,7 @@ func RecommendedTiDBCluster(ns, name string) *Recommendation {
 					},
 				},
 				TiKV: v1alpha1.TiKVSpec{
-					Replicas:             3,
+					Replicas:             *TiKVReplicas,
```
mahjonp: polish?

argo/workflow/bank.yaml
```diff
@@ -24,19 +24,26 @@ spec:
...
                   value: "20000"
                 - name: round
                   value: "100"
-
+                - name: loki-addr
+                  value: ""
+                - name: loki-username
+                  value: ""
+                - name: loki-password
+                  value: ""
```
yeya24: Is it better to change the loki default value back?

pkg/nemesis/iochaos.go
```diff
@@ -161,18 +206,19 @@ func (m ioMixed) template(
...
 	return m.template(ns, pods, chaosv1alpha1.OnePodMode,
-		chaosv1alpha1.FileSystemLayer, configName, "50ms", randomErrno, "50")
+		chaosv1alpha1.FileSystemLayer, configName, nil, "50ms", randomErrno, "50")
 }
 
 type ioChaos interface {
 	ioChaosType() chaosv1alpha1.IOChaosAction
 	template(ns string, pods []string, podMode chaosv1alpha1.PodMode,
-		layer chaosv1alpha1.IOLayer, configMap string, args ...string) chaosv1alpha1.IoChaosSpec
+		layer chaosv1alpha1.IOLayer, configMap string,methods []string, args ...string) chaosv1alpha1.IoChaosSpec
```
yeya24: A small nit. It is better to add a space between `configMap string,` and `methods []string`

pkg/test-infra/fixture/fixture.go
```diff
@@ -174,6 +175,7 @@ func init() {
 	flag.StringVar(&Context.TiDBConfigFile, "tidb-config", "", "path of tidb config file (cluster A in abtest case)")
 	flag.StringVar(&Context.TiKVConfigFile, "tikv-config", "", "path of tikv config file (cluster A in abtest case)")
 	flag.StringVar(&Context.PDConfigFile, "pd-config", "", "path of pd config file (cluster A in abtest case)")
+	flag.IntVar(&Context.TiKVReplicas, "tikv-replicas", 3, "number of tikv replicas")
```
yeya24: Do we also want to configure this parameter in abtest? If that, maybe we also need to add another parameter for the second cluster.

pkg/test-infra/tidb/recommend.go
yeya24: Here maybe better to use the same way like we do in `MaxFailoverCount`. Since now tikv replicas is configurable, maybe we need to make `MaxFailoverCount` configurable as well
``` Go
pointer.Int32Ptr(3)
```

pkg/nemesis/iochaos.go
fewdan: ok, i'll modify it.

pkg/nemesis/iochaos.go
```diff
@@ -86,12 +87,13 @@ func (d ioDelay) template(
...
 		Delay:      args[0],
 		Percent:    args[1],
+		Methods:    methods,
 	}
 }
 
 func (d ioDelay) defaultTemplate(ns, configMap string, pods []string) chaosv1alpha1.IoChaosSpec {
 	return d.template(ns, pods, chaosv1alpha1.OnePodMode,
-		chaosv1alpha1.FileSystemLayer, configMap, "50ms", "50")
+		chaosv1alpha1.FileSystemLayer, configMap, nil,"50ms", "50")
```
mahjonp:
```suggestion
		chaosv1alpha1.FileSystemLayer, configMap, nil, "50ms", "50")
```

pkg/test-infra/tidb/recommend.go
```diff
@@ -110,10 +107,10 @@ func RecommendedTiDBCluster(ns, name string) *Recommendation {
 					},
 				},
 				TiKV: v1alpha1.TiKVSpec{
-					Replicas:             *TiKVReplicas,
-					ResourceRequirements: fixture.WithStorage(fixture.Medium, "10Gi"),
+					Replicas:             int32(fixture.Context.TiKVReplicas),
+					ResourceRequirements: fixture.WithStorage(fixture.Medium, "100Gi"),
```
yeya24: Should this be `100Gi`? Or do we want this to be configurable?

argo/template/bank.yaml
```diff
@@ -43,9 +43,10 @@ spec:
...
             -case=bank \
             -client={{inputs.parameters.client}} \
+            -tikv-replicas={{inputs.parameters.tikv-replicas}} \
             -nemesis={{inputs.parameters.nemesis}} \
             -round={{inputs.parameters.round}} \
             -request-count={{inputs.parameters.request_count}} \
             -loki-addr={{inputs.parameters.loki-addr}} \
             -loki-username={{inputs.parameters.loki-username}} \
-            -loki-password={{inputs.parameters.loki-password}}
+            -loki-password={{inputs.parameters.loki-password}}
```
yeya24: Needs a new line here

argo/workflow/bank.yaml
```diff
@@ -56,5 +56,4 @@ spec:
                 - name: request_count
                   value: "{{workflow.parameters.request_count}}"
                 - name: round
-                  value: "{{workflow.parameters.round}}"
-
+                  value: "{{workflow.parameters.round}}"
```
yeya24: ditto

pkg/test-infra/tidb/recommend.go
mahjonp: Yes, we can make it configurable on another PR.

PR #135, opened 2020-03-25
Add TiDB_row_format_version and Update setup

fewdan:

Add TiDB_row_format_version and Update setup

Add TiDB_row_format_version and Update setup

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


mahjonp:
You need consider below two points:
1) how to make it easily configurable? we have many configurations use global variable to set, how to make it common use?
2) how to make A/B test, eg, set default on cluster one and set another value on cluster two.

PR #143, opened 2020-03-27
Support tiflash

yeya24:
Add tiflash support. Since this PR is kind of large, I will open another PR for chaos for tiflash

PR #147, opened 2020-03-27
Add tiflash argo workflow

yeya24:
Please merge this after https://github.com/pingcap/tipocket/pull/143 get merged

### DIFF
--- a/argo/workflow/tiflash.yaml
+++ b/argo/workflow/tiflash.yaml
@@ -1,0 +1,67 @@
+metadata:
+  generateName: tipocket-tiflash-
+  namespace: argo
+spec:
+  entrypoint: call-tipocket-tiflash
+  onExit: call-export-logs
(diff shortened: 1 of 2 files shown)


argo/workflow/tiflash.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: image_version
+        value: release-4.0-nightly
+      - name: storage_class
+        value: pd-ssd
+      - name: nemesis
+        value: ""
+      - name: run_time
+        value: "6h"
+      - name: tiflash_image
+        value: "yeya24/tics:4.0"
```
zhouqiang-cl: Can we use official image

yeya24: No. The official has some problems in its image. So I have to build my own. I think it is okay to update it later

PR #151, opened 2020-03-29
Update cdc

yeya24:

1. Add a headless service to CDC statefulset. This service exposes the status/metrics of CDC instances, which is useful for debugging and further monitoring.
2. Make the labels consistent with other components.
3. Use "k8s.io/apimachinery/pkg/api/errors" to check errors instead of "github.com/pingcap/errors"

### DIFF
--- a/pkg/test-infra/cdc/operation.go
+++ b/pkg/test-infra/cdc/operation.go
@@ -21,6 +21,8 @@ import (
 	_ "github.com/go-sql-driver/mysql" // mysql driver
 	"github.com/ngaut/log"
 	"github.com/pingcap/errors"
+	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


pkg/test-infra/cdc/operation.go
```diff
@@ -84,6 +89,9 @@ func (c *CDCOps) Delete(tc *Recommendation) error {
 }
 
 func (c *CDCOps) ApplyCDC(cc *CDC) error {
+	if err := c.applyObject(cc.Service); err != nil {
```
cwen0: why applying CDC job before creating CDC object?

yeya24: Why do you mean `CDC job` here? I apply that job after this function

cwen0: my fault.

PR #152, opened 2020-03-30
add v_bank case

coocood:

Add an extensible v_bank case

Add an extensible v_bank case

### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

Related changes

```release-note
add v_bank case
```


cmd/util/suit.go
```diff
@@ -301,3 +301,27 @@ func nemesisGeneratorNames(gens []core.NemesisGenerator) (names []string) {
...
+	IP   string
+	Port int32
+}
+
+// NewLocalProvisioner creates a LocalProvisioner
+func NewLocalProvisioner(ip string, port int32) *LocalProvisioner {
+	return &LocalProvisioner{IP: ip, Port: port}
+}
+
+// TearDown implements the Provisioner interface.
```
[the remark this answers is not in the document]
mahjonp: @coocood I added a `LocalClusterProvisioner` before, you could use it https://github.com/pingcap/tipocket/blob/46f18d3af7dd46884be45e87fc0cff1f1c099090/pkg/cluster/local_cluster.go#L27

mahjonp:
I added a `LocalClusterProvisioner` on https://github.com/pingcap/tipocket/pull/151 before, you can use it on debug.

tests/vbank/v_bank.go
```diff
@@ -0,0 +1,839 @@
...
+	CreateResult   *CreateResult
+	DeleteResult   *DeleteResult
+	TransferResult *TransferResult
+	OK             bool
+	Err            string
+}
+
+// IsUnknown implements UnknownResponse interface
+func (r *Response) IsUnknown() bool {
+	return false
```
mahjonp: If always return false, how to resolve the situation when client doesn't known whether the request is successful?

coocood: By checking the transaction status in `v_bank_txn_status` table.

PR #156, opened 2020-03-30
Add tiflash abtest

yeya24:


yeya24:
Ready for review now.

pkg/pocket/executor/common.go
```diff
@@ -116,3 +125,37 @@ func (e *Executor) TxnRollback() error {
...
+	sql := fmt.Sprintf("SELECT AVAILABLE FROM information_schema.tiflash_replica WHERE TABLE_SCHEMA = '%s' and TABLE_NAME = '%s'",
+		e.dbname, table)
+	for {
+		res, err := e.GetConn().Select(sql)
+		if err != nil {
+			return err
+		}
+		// table doesn't exist
+		if res == nil {
+			return nil
```
you06, line 145: If the table does not exist, why not wait and check it next time?

pkg/pocket/executor/common.go
```diff
@@ -116,3 +125,37 @@ func (e *Executor) TxnRollback() error {
...
+// since the tiflash cluster is always conn1 in TiFlash abtest, so it is
+// okay to just use single test ddl here.
+func (e *Executor) createTiFlashTableReplica(table string) error {
+	return e.SingleTestExecDDL(fmt.Sprintf("ALTER TABLE %s SET TIFLASH REPLICA 1", table))
+}
+
+func (e *Executor) WaitTiFlashTableSync(table string) error {
+	sql := fmt.Sprintf("SELECT AVAILABLE FROM information_schema.tiflash_replica WHERE TABLE_SCHEMA = '%s' and TABLE_NAME = '%s'",
+		e.dbname, table)
+	for {
```
you06, line 138: May add a timeout for the sync function

pkg/pocket/core/generate.go
```diff
@@ -50,8 +49,12 @@ func (c *Core) beforeGenerate() error {
 		}
 		log.Infof("table %s generate", sql.SQLTable)
 
-		if c.cfg.Mode == "tiflash" {
-			c.createTiFlashTableReplica(sql, e)
+		if e.TiFlash {
+			if err := e.WaitTiFlashTableSync(sql.SQLTable); err != nil {
```
you06, line 53: Can we waiting for table sync together after all create table statements are executed?

yeya24: Good idea. Will update it tomorrow

pkg/pocket/executor/common.go
yeya24: This is for the case when the generated table name doesn't exist like `s4`, `s5`. Seems they won't be created. So I have to return nil. To be honest, I'm not really sure about this case.

mahjonp, line 145: if table doesn't exist, is returning nil ok?

pkg/pocket/executor/abtest.go
```diff
@@ -147,10 +160,13 @@ func (e *Executor) abTestSelect(sql string) error {
 		err2 error
 	)
 	wg.Add(2)
-	go func() {
+	go func(sql string) {
+		if e.TiFlash {
+			sql = setIsolationEngine + sql
```
you06, line 165: `setIsolationEngine` set a session variable, it does not need to be executed everytime.

pkg/pocket/executor/common.go
you06: `s4`, `s5` will only be generated in select statements by mistake. I find you wait for table sync before read statements, why not just wait for the sync after create table statements?

you06:
rest LGTM

pkg/pocket/executor/common.go
yeya24: If I don't wait before read, it will produce some error.

yeya24: Not sure. This is the case for `s4` and `s5`, I am not really sure how to deal with that. Since these tables are not really creating, so it will always be nil.

yeya24:
I am thinking about working on this again after #162 is merged. Then I can clean up some code in this PR.

yeya24:
This pr is updated

yeya24:
I don't want to delay too much of the tiflash + cdc case, but it relies on this PR gets merged first. So I will add that feature into this PR together. Hope you don't mind it.

pkg/pocket/core/execute.go
```diff
@@ -60,14 +60,14 @@ func (c *Core) executeByID(id int, sql *types.SQL) {
 // and the manipulated db should be c.dbname
 func (c *Core) coreInitDatabaseExecute(sql *types.SQL) error {
 	switch c.cfg.Mode {
-	case "single":
+	case "single", "tiflash":
 		return errors.Trace(c.coreExec.GetConn().Exec(sql.SQLStmt))
-	case "binlog", "tiflash":
+	case "binlog", "tiflash-binlog":
```
mahjonp, line 65: what's the meaning of tiflash-binlog?

you06: I think it's for CDC, binlog means write from upstream and compare it with downstream.

yeya24: Just apply the same naming pattern for the cdc case. Since it uses `binlog` name as well

PR #168, opened 2020-04-02
control: refactor nemesis logs

mahjonp:

Add more logs on nemesis and will retry when recovery from nemesis fails

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/control/control.go
```diff
@@ -245,23 +244,32 @@ func (c *Controller) RunSelfScheduled() {
...
+		})
+	}
+	if err := g.Wait(); err != nil {
+		log.Fatalf("run client error, %+v", errors.ErrorStack(err))
 	}
 
-	ncancel()
+	// cancel nCtx and wait nemesis ended
+	nCancel()
+	nemesisWg.Wait()
```
you06, line 271: Seems this wait group is equal to a channel

pkg/control/control.go
```diff
@@ -378,61 +383,68 @@ LOOP:
...
 			default:
 			}
-
-			log.Printf("begin to run %s nemesis generator", g.Name())
-			ops := g.Generate(c.cfg.Nodes)
-
-			wg.Add(len(ops))
+			var (
+				ops = g.Generate(c.cfg.Nodes)
+				g   errgroup.Group
```
you06, line 388: It's better to use another variable name instead of overwriting `g`

pkg/control/control.go
mahjonp: I found that we lost this wait statement in preivous.

pkg/control/control.go
mahjonp: Good catch, I didn't notice that.

mahjonp: Updated.

pkg/control/control.go
you06: I just feel a little odd about wait group has only one task (it's equal to waiting for a channel), but it's ok to me.

PR #180, opened 2020-04-08
Refine the read-stress test by calculating average latency in the last minute

sticnarf:

The read-stress test may be too easy to fail because once there is a jitter in a small query, it will exceed the limit for small queries.

Calculate the average latency in the past minute and compare it to the limit. It also prints the 99th percentile of the latency.


sticnarf:
PTAL @mahjonp 
Actually the previous version can work on my own laptop. I have no idea why it fails on K8S. But this PR adds more information even on failure.

PR #187, opened 2020-04-14
retry when abtest compare occurred an error

mahjonp:

When startCheckConsistency occurred an error, it will fatal immediately, that't not what we expect.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/pocket/core/check.go
+++ b/pkg/pocket/core/check.go
@@ -52,9 +52,8 @@ func (c *Core) startCheckConsistency() {
 					// TODO: pass error by channel, stop process from outside
 					log.Errorf("compare data error %+v", errors.ErrorStack(err))
 					return false, nil
-				} else {
-					return true, nil
 				}
+				return true, nil
 			})
 			if err != nil {
 				log.Fatalf("compare data failed %v", err)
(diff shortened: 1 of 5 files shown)


pkg/pocket/core/check.go
```diff
@@ -39,11 +40,24 @@ func (c *Core) startCheckConsistency() {
...
 		go func(t int) {
+			var (
+				result bool
+				err    error
+			)
 			log.Info("ready to compare data")
-			result, err := c.checkConsistency(false)
+
+			err = wait.PollImmediate(1*time.Minute, 10*time.Minute, func() (bool, error) {
+				result, err = c.checkConsistency(false)
```
you06, line 50: This error can be caused by database unavailable either data inconsistency, and retry is necessary is for the first condition only.

you06:
Besides, the compare process will stop execution and waiting for sync first, move the error retry into this function would be better.
https://github.com/pingcap/tipocket/blob/8dfba30f71eebe6eb7aa77c38a6b8a4fc0447f5d/pkg/pocket/core/check.go#L249-L259

pkg/pocket/core/check.go
mahjonp: Seems there will be a panic when data inconsistency occurred?

mahjonp:
> Besides, the compare process will stop execution and waiting for sync first, move the error retry into this function would be better.
> https://github.com/pingcap/tipocket/blob/8dfba30f71eebe6eb7aa77c38a6b8a4fc0447f5d/pkg/pocket/core/check.go#L249-L259

Yes, that's better. I will move retrying here.

mahjonp:
> Besides, the compare process will stop execution and waiting for sync first, move the error retry into this function would be better.
> https://github.com/pingcap/tipocket/blob/8dfba30f71eebe6eb7aa77c38a6b8a4fc0447f5d/pkg/pocket/core/check.go#L249-L259

The outer retry is required, it's used to catch the below error before compareData
https://github.com/pingcap/tipocket/blob/8dfba30f71eebe6eb7aa77c38a6b8a4fc0447f5d/pkg/pocket/core/check.go#L169-L172

pkg/pocket/core/check.go
you06: Yes, so there should be only inconsistency and time exceeded errors.

mapleFU: Why not add a cause checking here?

PR #188, opened 2020-04-14
update graph.go

fewdan:
update graph.go

update graph.go

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```

### DIFF
--- a/pkg/elle/core/graph.go
+++ b/pkg/elle/core/graph.go
@@ -196,11 +196,73 @@ func (g *DirectedGraph) FilterRelationships(rels []Rel) *DirectedGraph {
 	return fg
 }
 
+// BfsOut searches from a vertices set, returns all vertices searchable
+// search to downstream vertices from `out` edges
+func (g *DirectedGraph) BfsOut(initV []Vertex) []Vertex {
+	queue := initV
+	V := []Vertex{}
+	have_visited := make(map[Vertex]bool)
+	for _, vertex := range initV {
+		have_visited[vertex] = true
+		V = append(V, vertex)
+	}
+
+	for len(queue) > 0 {
+		cur := queue[0]
+		if have_visited[cur] == false {
+			have_visited[cur] = true
+			V = append(V, cur)
+		}
+		queue = queue[1:]
+
+		for next, _ := range g.Outs[cur] {
+			if have_visited[next] == true {
+				continue
+			}
+			queue = append(queue, next)
+		}
+	}
+	return V
+}
+
+// BfsIn searches from a vertices set, returns all vertices searchable
+// search to upstream vertices from `in` edges
+func (g *DirectedGraph) BfsIn(initV []Vertex) []Vertex {
+	queue := initV
+	V := []Vertex{}
+	have_visited := make(map[Vertex]bool)
+	for _, vertex := range initV {
+		have_visited[vertex] = true
+		V = append(V, vertex)
+	}
+
+	for len(queue) > 0 {
+		cur := queue[0]
+		if have_visited[cur] == false {
+			have_visited[cur] = true
+			V = append(V, cur)
+		}
+		queue = queue[1:]
+		
+		for _, next := range g.Ins[cur] {
+			if have_visited[next] == true {
+				continue
+			}
+			queue = append(queue, next)
+		}
+	}
+	return V
+}
+
 // Bfs searches from a vertices set, returns all vertices searchable
 // out = true means search to downstream vertices from `out` edges
 // out = false means search to upstream vertices from `in` edges
 func (g *DirectedGraph) Bfs(initV []Vertex, out bool) []Vertex {
-	panic("impl me")
+	if out == true {
+		return g.BfsOut(initV)
+	} else {
+		return g.BfsIn(initV)
+	}
 }
 
 // SCC indexes all vertices of a strongly connected component
--- a/pkg/elle/core/graph.go
+++ b/pkg/elle/core/graph.go
@@ -20,18 +20,9 @@ type DirectedGraph struct {
 // Vertices returns the set of all vertices in graph
 func (g *DirectedGraph) Vertices() []Vertex {
 	vertices := []Vertex{}
-	haveIn := make(map[Vertex]bool)
-
-	for key := range g.Ins {
-		vertices = append(vertices, key)
-		haveIn[key] = true
-	}
 
 	for key := range g.Outs {
-		_, ok := haveIn[key]
-		if !ok {
-			vertices = append(vertices, key)
-		}
+		vertices = append(vertices, key)
 	}
 	return vertices
 }
@@ -91,6 +82,19 @@ func (g *DirectedGraph) Link(v Vertex, succ Vertex, rel Rel) {
 	if !ok {
 		g.Outs[v] = make(map[Vertex][]Rel)
 	}
+	_, ok = g.Outs[succ]
+	if !ok {
+		g.Outs[succ] = make(map[Vertex][]Rel)
+	}
+
+	_, ok = g.Ins[v]
+	if !ok {
+		g.Ins[v] = []Vertex{}
+	}
+	_, ok = g.Ins[succ]
+	if !ok {
+		g.Ins[succ] = []Vertex{}
+	}
 
 	haveRel := false
 	for _, item := range g.Outs[v][succ] {
@@ -131,18 +135,12 @@ func (g *DirectedGraph) LinkAllTo(xs []Vertex, y Vertex, rel Rel) {
 
 func (g *DirectedGraph) UnLink(a, b Vertex) {
 	delete(g.Outs[a], b)
-	if len(g.Outs[a]) == 0 {
-		delete(g.Outs, a)
-	}
 
 	for id , vertex := range g.Ins[b] {
 		if vertex == a {
 			g.Ins[b] = append(g.Ins[b][:id], g.Ins[b][id+1:]...)
 			break
 		}
-	}
-	if len(g.Ins[b]) == 0 {
-		delete(g.Ins, b)
 	}
 }
 
@@ -377,21 +375,24 @@ func (g *DirectedGraph) StronglyConnectedComponents() []SCC {
 // MapVertices takes a function of vertices, returns graph with all
 //  vertices mapped via f
 func (g *DirectedGraph) MapVertices(f func(interface{}) interface{}) *DirectedGraph {
-	fg := g.Fork()
-	vertices := fg.Vertices()
+	var fg DirectedGraph
+	fg.Ins = make(map[Vertex][]Vertex)
+	fg.Outs = make(map[Vertex]map[Vertex][]Rel)
+
+	vertices := g.Vertices()
 	for _, x := range vertices {
 		_, ok := g.Outs[x]
 		if ok == true {
 			for y, _:= range g.Outs[x] {
 				rels := g.Outs[x][y]
-				fg.UnLink(x, y)
 				for _, item := range rels {
 					fg.Link(Vertex{f(x.Value)}, Vertex{f(y.Value)}, item)
 				}
 			}
 		}
 	}
-	return fg
+
+	return &fg
 }
 
 // RenumberGraph takes a Graph and rewrites each vertex to a unique integer, returning the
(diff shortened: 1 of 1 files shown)


pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
 
 // Vertices returns the set of all vertices in graph
 func (g *DirectedGraph) Vertices() []Vertex {
-	panic("impl me")
+	vertices := []Vertex{}
+	have_in := make(map[Vertex]bool)
+
+	for key := range g.Ins {
```
mahjonp, line 25: Iterating just g.Outs is enough, and have_in should be haveIn

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
+	return vertices
 }
 
-// In returns inbound edges to v in graph g
+// In returns inbound vertices to v in graph g
 func (g *DirectedGraph) In(v Vertex) []Vertex {
-	panic("impl me")
+	_, ok := g.Ins[v]
+	if !ok {
+		return nil
```
mahjonp, line 43: returning an empty slice is better here.

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
+	}
 }
 
-// Out returns outbound edges from v in graph g
+// Out returns outbound vertices from v in graph g
 func (g *DirectedGraph) Out(v Vertex) []Vertex {
-	panic("impl me")
+	_, ok := g.Outs[v]
+	if !ok {
+		return nil
```
mahjonp, line 53: ditto.

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
-func (g *DirectedGraph) Link(v Vertex, succ []Vertex, rel Rel) {
-	panic("impl me")
+func (g *DirectedGraph) Link(v Vertex, succ Vertex, rel Rel) {
+
+	_, ok := g.Outs[v]
+	if !ok {
+		g.Outs[v] = make(map[Vertex][]Rel)
+	}
+
+	g.Outs[v][succ] = append(g.Outs[v][succ], rel)
```
mahjonp, line 96: If rel is already in the g.Outs[v][succ], we'd better ignore this append here.

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
+	g.Ins[succ] = append(g.Ins[succ], v)
 }
 
 // LinkToAll links x to all ys
 func (g *DirectedGraph) LinkToAll(x Vertex, ys []Vertex, rel Rel) {
-	panic("impl me")
+	for _, y := range ys {
```
mahjonp, line 102: How about reuse Link?

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
+		g.Outs[x][y] = append(g.Outs[x][y], rel)
+		g.Ins[y] = append(g.Ins[y], x)
+	}
 }
 
 // LinkAllTo links all xs to y
-func (g *DirectedGraph) LinkAllTo(xs []Vertex, y Vertex) {
-	panic("impl me")
+func (g *DirectedGraph) LinkAllTo(xs []Vertex, y Vertex, rel Rel) {
+	for _, x := range xs {
```
mahjonp, line 116: ditto.

pkg/elle/core/graph.go
```diff
@@ -19,41 +19,128 @@ type DirectedGraph struct {
...
+		}
+
+		g.Outs[x][y] = append(g.Outs[x][y], rel)
+		g.Ins[y] = append(g.Ins[y], x)
+	}
 }
 
 func (g *DirectedGraph) UnLink(a, b Vertex) {
-	panic("impl me")
+	delete(g.Outs[a], b)
```
mahjonp, line 129: polish this function?

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,250 @@ type DirectedGraph struct {
...
+	}
 }
 
-// Out returns outbound edges from v in graph g
+// Out returns outbound vertices from v in graph g
 func (g *DirectedGraph) Out(v Vertex) []Vertex {
-	panic("impl me")
+	_, ok := g.Outs[v]
+	if !ok {
+		return []Vertex{}
```
you06, line 44:
```suggestion
		return nil
```

pkg/elle/core/graph.go
```diff
@@ -122,21 +131,27 @@ func (g *DirectedGraph) LinkAllTo(xs []Vertex, y Vertex, rel Rel) {
 
 func (g *DirectedGraph) UnLink(a, b Vertex) {
 	delete(g.Outs[a], b)
+	if len(g.Outs[a]) == 0 {
+		delete(g.Outs, a)
```
mahjonp, line 135: Why delete a vertex?

pkg/elle/core/graph.go
```diff
@@ -122,21 +131,27 @@ func (g *DirectedGraph) LinkAllTo(xs []Vertex, y Vertex, rel Rel) {
...
+	}
 
 	for id , vertex := range g.Ins[b] {
 		if vertex == a {
 			g.Ins[b] = append(g.Ins[b][:id], g.Ins[b][id+1:]...)
 			break
 		}
 	}
+	if len(g.Ins[b]) == 0 {
```
mahjonp, line 144: ditto

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+	return vertices
 }
 
-// In returns inbound edges to v in graph g
+// In returns inbound vertices to v in graph g
 func (g *DirectedGraph) In(v Vertex) []Vertex {
-	panic("impl me")
+	_, ok := g.Ins[v]
+	if !ok {
+		return []Vertex{}
```
mahjonp, line 34: So, will we meet the situation that v isn't in `Ins`?

mapleFU: `return nil` is posible.

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+	}
+	_, ok = g.Ins[succ]
+	if !ok {
+		g.Ins[succ] = []Vertex{}
+	}
+
+	haveRel := false
+	for _, item := range g.Outs[v][succ] {
+		if item == rel {
+			haveRel = true
```
mahjonp, line 102: IMO you don't need haveRel, we can return if `rel` already is in g.Outs[v][succ].

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
-func (g *DirectedGraph) RemoveRelationship(rel Rel) *DirectedGraph {
-	panic("impl me")
+// IntersectionRel returns the union of res1 and res2
+func IntersectionRel(rels1 []Rel, rels2 []Rel) []Rel{
+	rel := []Rel{}
+	for _, rel1 := range rels1 {
+		both := false
+		for _, rel2 := range rels2{
+			if rel1 == rel2 {
+				both = true
```
mahjonp, line 158:
```suggestion
				rel = append(rel, rel1)
```

don't need the `both` variable.

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+
+	_, ok = g.Ins[v]
+	if !ok {
+		g.Ins[v] = []Vertex{}
+	}
+	_, ok = g.Ins[succ]
+	if !ok {
+		g.Ins[succ] = []Vertex{}
+	}
+
```
mapleFU, line 98: https://paste.ubuntu.com/p/fM69YNN6sq/

Actually I don't think you need to initialize it.

mahjonp: No, we need ensure all keys at `Outs` are also at `Ints` and vice versa. @mapleFU

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+			rel = append(rel, rel1)
+		}
+	}
+	return rel
 }
 
 // ProjectRelationship filters a graph to just those edges with the given relationship
 func (g *DirectedGraph) ProjectRelationship(rel Rel) *DirectedGraph {
-	panic("impl me")
+	fg := g.Fork()
```
mahjonp, line 171: `ProjectRelationship` can reuse `FilterRelationships`  method

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+		vertices = append(vertices, key)
+	}
+	return vertices
 }
 
 // Edges returns the edge between two vertices
 func (g *DirectedGraph) Edges(a, b Vertex) []Edge {
-	panic("impl me")
+	_, oka := g.Outs[a]
+	_, okb := g.Ins[b]
```
you06, line 58: I think you only need to check `g.Outs[a]` and `g.Outs[a][b]` exist, `g.Ins` is used for look for reverse paths.

pkg/elle/core/graph.go
```diff
@@ -97,20 +285,143 @@ type SCC struct {
...
+	}
+	return scc
 }
 
 // MapVertices takes a function of vertices, returns graph with all
 //  vertices mapped via f
 func (g *DirectedGraph) MapVertices(f func(interface{}) interface{}) *DirectedGraph {
-	panic("impl me")
+	var fg DirectedGraph
+	fg.Ins = make(map[Vertex][]Vertex)
```
mahjonp, line 379: extract to a new method?

pkg/elle/core/graph.go
you06: Does this function add an edge from `v` to `succ`? If so, what's the purpose of creating `g.Outs[succ]` and `g.Ins[v]`?

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
+			if rel1 == rel2 {
+				both = true
+				break
+			}
+		}
+		if both == true {
+			rel = append(rel, rel1)
+		}
+	}
+	return rel
```
you06, line 166: https://play.golang.org/p/SLIqRTgvpfW What about this?

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
 func (g *DirectedGraph) FilterRelationships(rels []Rel) *DirectedGraph {
-	panic("impl me")
+	fg := g.Fork()
+
+	vertices := fg.Vertices()
+	for _, x := range vertices {
+		_, ok := g.Outs[x]
+		if ok == true {
+			for y, _:= range g.Outs[x] {
+				rels := IntersectionRel(g.Outs[x][y], rels)
```
you06, line 200: It seems you add the input `rels` into every edges of the graph?

pkg/elle/core/graph.go
```diff
@@ -19,75 +19,263 @@ type DirectedGraph struct {
...
-// keepEdgeValues transforms a graph by applying a function (f edge-value) to each edge in the
-//  graph. Where the function returns `nil`, removes that edge altogether
-func (g *DirectedGraph) keepEdgeValues(f func(interface{}) interface{}) *DirectedGraph {
-	panic("impl me")
-}
-
-// RemoveRelationship removes the given relationship from it
-func (g *DirectedGraph) RemoveRelationship(rel Rel) *DirectedGraph {
-	panic("impl me")
+// IntersectionRel returns the union of res1 and res2
```
mahjonp, line 151:
```suggestion
// IntersectionRel returns the intersection of res1 and res2
```

pkg/elle/core/graph.go
fewdan: If the vertices are normal, this situation will not be encountered, otherwise it will return nil.

pkg/elle/core/graph.go
fewdan: If there is no initialization, the points without in-degrees are not in "Ins". Similarly, the points without out-degrees are not in "Outs". In this case, we need to modify "Vertices()".

PR #192, opened 2020-04-15
Implement elle core

you06:

Implement elle's core module.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

### DIFF
--- a/pkg/elle/core/core.go
+++ b/pkg/elle/core/core.go
@@ -1,7 +1,9 @@
 package core
 
+// Rel stands for relation in dependencies
 type Rel string
 
+// Rel enums
 const (
 	Empty    Rel = ""
 	WW       Rel = "ww"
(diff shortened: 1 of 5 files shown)


pkg/elle/core/core.go
```diff
@@ -54,22 +79,142 @@ func (c *CycleExplainer) renderCycleExplanation(explainer DataExplainer, circle
 	panic("impl me")
 }
 
-//  A function which takes a history and returns a {:graph, :explainer, :anomalies} map; e.g. realtime-graph.
-type Analyzer func(history History) (Anomalies, DirectedGraph, DataExplainer)
+// Analyzer is a function which takes a history and returns a {:graph, :explainer} map; e.g. realtime-graph.
+type Analyzer func(history History) (DirectedGraph, DataExplainer)
```
mahjonp, line 83: Perhaps we need retain `Anomalies`, it's used on `Combine` func and rw_register workload.

pkg/elle/core/core.go
```diff
@@ -54,22 +79,142 @@ func (c *CycleExplainer) renderCycleExplanation(explainer DataExplainer, circle
...
-	panic("implement me")
+func ProcessGraph(history History) (DirectedGraph, DataExplainer) {
+	var (
+		okHistory = history.FilterType(OpTypeOk)
+		processes map[int]struct{}
+		graphs    []DirectedGraph
+	)
+
+	for _, op := range okHistory {
+		if op.Process != 0 {
```
mahjonp, line 132: Why filter out Process 0? what's Process 0 means?

you06: Process should not be 0 and 0 will be the default value when `Op` struct inits. Filter out Process with value 0 to exclude the operations which don't have a Process value.

mahjonp: How about change Process type to *int?

PR #193, opened 2020-04-16
Add some codes for explainer

mapleFU:
Signed-off-by: mapleFU <1506118561@qq.com>

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/elle/core/core.go
+++ b/pkg/elle/core/core.go
@@ -11,12 +11,12 @@ const (
 	Realtime Rel = "realtime"
 )
 
-type DependType int
+type DependType string
 
 const (
-	RealtimeDepend DependType = iota
-	MonotonicDepend
-	ProcessDepend
+	RealtimeDepend  DependType = "realtime"
+	MonotonicDepend DependType = "monotonic"
+	ProcessDepend   DependType = "process"
 )
 
 type ExplainResult interface {
@@ -46,7 +46,8 @@ func CombineExplainer([]DataExplainer) DataExplainer {
 	panic("implement me")
 }
 
-type PathType = interface{}
+// TODO: we may refine it later
+type PathType = Op
 
 type Circle struct {
 	// Eg. [2, 1, 2] means a circle: 2 -> 1 -> 2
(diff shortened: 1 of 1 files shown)


pkg/elle/core/core.go
```diff
@@ -11,6 +11,18 @@ const (
 	Realtime Rel = "realtime"
 )
 
+type DependType int
+
+const (
+	RealtimeDepend DependType = iota
```
mahjonp, line 17: Do not use iota and int type, it's hard to read.

pkg/elle/core/core.go
```diff
@@ -24,7 +36,7 @@ func MergeAnomalies(anomalies ...Anomalies) Anomalies {
 type DataExplainer interface {
 	// Given a pair of operations a and b, explains why b depends on a, in the
 	//    form of a data structure. Returns `nil` if b does not depend on a.
-	ExplainPairData() interface{}
+	ExplainPairData(p1, p2 PathType) ExplainResult
```
mahjonp, line 39: PathType seems always be core.Op type? IMO we can use core.Op here directly.

pkg/elle/core/core.go
```diff
@@ -34,23 +46,41 @@ func CombineExplainer([]DataExplainer) DataExplainer {
...
+	Path []PathType
 }
 
 // TODO: refine me
-type Step struct{}
+type Step struct {
+	Result ExplainResult
+}
+
+func explainCyclePairData(pairExplainer DataExplainer, prePath PathType, currentPath PathType) ExplainResult {
```
mahjonp, line 61: ditto. use core.Ops may better.

pkg/elle/core/core.go
mapleFU: Emm what is better here?

mahjonp: Perhaps the string literal is better

pkg/elle/core/core.go
mapleFU: OK

PR #196, opened 2020-04-16
replace current tiflash with tiflash cr

yeya24:


pkg/test-infra/tidb/operation.go
```diff
@@ -63,8 +63,8 @@ type Ops struct {
 }
 
 // New ...
-func New(namespace, name string, config fixture.TiDBClusterConfig) *Ops {
-	return &Ops{cli: tests.TestClient.Cli, tc: RecommendedTiDBCluster(namespace, name, config),
+func New(namespace, name string, config fixture.TiDBClusterConfig, enableTiFlash bool) *Ops {
```
mahjonp, line 66: How about putting enableTiFlash into fixture.TiDBClusterConfig and using TiFlashReplicas to control whether or not require tiflash?

yeya24: Yes, that is a good idea

yeya24:
Updated

pkg/test-infra/tidb/recommend.go
```diff
@@ -220,11 +249,23 @@ func RecommendedTiDBCluster(ns, name string, clusterConfig fixture.TiDBClusterCo
...
 			},
 		},
 	}
 
+	if clusterConfig.TiFlashReplicas > 0 {
+		r.EnableTiFlash(clusterConfig.TiFlashImageVersion, clusterConfig.TiFlashReplicas)
+
+		// TODO(yeya24): There is a bug in the operator side so we can only set it
+		// manually https://github.com/pingcap/tidb-operator/issues/2219.
+		// Remove this after it fixes the bug.
```
mahjonp, line 261: So it's ok to use tidb-operator with TiFlash now? I see DanielZhangQD said that it's still in progress?

yeya24: For deploying tiflash it is okay now. WIP is meant for scale in and scale out, which is not necessary in tipocket

PR #202, opened 2020-04-23
add resolve-lock test

youjiali1995:
add resolve-lock test for green gc


jackysp:
Should we merge it to master?

youjiali1995:
Waiting for https://internal.pingcap.net/jira/browse/ONCALL-1291 merged and it needs a special docker image.

youjiali1995:
@mahjonp @MyonKeminta PTAL

argo/template/resolve-lock.yaml
```diff
@@ -0,0 +1,62 @@
...
+          - name: pd_config
+            default: ""
+          - name: tikv-replicas
+            default: "3"
+      metadata:
+        labels:
+          ns: "{{inputs.parameters.ns}}"
+      container:
+        name: tipocket
+        image: 'mahjonp/tipocket:green_gc'
```
mahjonp, line 42: revert pingcap/tipocket

And could you update template and workflow, we have updated them many times since this PR.

youjiali1995: Ahh, this case needs a special image that enables failpoints in TiDB.

argo/template/resolve-lock.yaml
```diff
@@ -8,55 +8,65 @@ spec:
...
             -nemesis={{inputs.parameters.nemesis}} \
-            -run-time={{inputs.parameters.run_time}} \
+            -run-time={{inputs.parameters.run-time}} \
+            -tikv-replicas={{inputs.parameters.tikv-replicas}} \
             -loki-addr={{inputs.parameters.loki-addr}} \
             -loki-username={{inputs.parameters.loki-username}} \
-            -loki-password={{inputs.parameters.loki-password}}
\ No newline at end of file
+            -loki-password={{inputs.parameters.loki-password}}
+            -plugin.leak.eat=http://pingcap-dev.hk.ufileos.com/schrodinger/eat.txt
```
mahjonp, line 72: This line can be removed safely.

argo/template/resolve-lock.yaml
mahjonp: > Ahh, this case needs a special image that enables failpoints in TiDB.

Perhaps you can ask @zyguan for help, they may have the CI job

zyguan: You can use `hub.pingcap.net/qa/{tidb,tikv,pd}:{master,release-4.0}-failpoint` directly, they are built daily.

However, since a special env var needed to be set for enabling the failpoint, we need to use a custom startup script. see https://github.com/pingcap/automated-tests/blob/dev/pkg/resource/impl/k8s/assets/start-tidb.sh#L40

mahjonp: @zyguan does tidb-server support to enable failpoint in runtime? For example activate some failpoints through HTTP endpoint?

zyguan: Failpoint itself serves on [`GO_FAILPOINTS_HTTP`](https://github.com/pingcap/failpoint/blob/b230c36c413c4574ab97d09419f7e796eec4e565/failpoints.go#L76-L81), it also requires specifying a env var when start tidb. 

Besides, the `enableTestAPI` is special, one should enable it before tidb setup status service. So, AFIK, there is no workaround currently.

PR #207, opened 2020-04-24
cdc-tipocket: support to enable kafka sink

5kbpers:
Signed-off-by: 5kbpers <tangminghua@pingcap.com>

Support to enable kafka sink for cdc.

Add an option for enabling kafka sink.

### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

No.


pkg/test-infra/cdc/recommend.go
```diff
@@ -163,6 +170,147 @@ func newCDC(ns, name string) *CDC {
...
+			Spec: batchv1.JobSpec{
+				Template: corev1.PodTemplateSpec{
+					Spec: corev1.PodSpec{
+						Containers: []corev1.Container{
+							{
+								Name:            "cdc-kafka-consumer",
+								Image:           buildCDCImage("cdc-kafka-consumer"),
+								ImagePullPolicy: corev1.PullAlways,
+								Command: []string{
+									"/cdc_kafka_consumer",
```
mahjonp, line 300: What's the cdc_kafka_consumer  function ?

mahjonp:
https://github.com/pingcap/tipocket/blob/13d9da606eebbe76f3a38122e24d04555d5ef753/pkg/test-infra/clusters.go#L146-L155

We use the composition way to build a cluster, so downstream tidb cluster will also be deployed if you set EnableKafka now. You should also change here.

pkg/test-infra/cdc/recommend.go
5kbpers: Pull data from kafka, decode and sink them to downstream tidb.

5kbpers:
@mahjonp Actually we need downstream tidb.

PR #212, opened 2020-04-28
Refactor logs download

yeya24:
Refactor the current log download logic.
Will download from loki every 4 minute


cmd/util/suit.go
```diff
@@ -79,16 +82,20 @@ func (suit *Suit) Run(ctx context.Context) {
...
 		suit.NemesisGens,
 		suit.ClientRequestGen,
 		suit.VerifySuit,
-		loki.NewClient(fixture.Context.LokiAddress,
+		loki.NewClient(startTime, fixture.Context.LokiAddress,
 			fixture.Context.LokiUsername, fixture.Context.LokiPassword),
+		fixture.Context.LogPath,
 	)
+	if err != nil {
+		log.Fatalf("failed to create log directory %s", fixture.Context.LogPath)
```
mahjonp, line 97: put panic into inner?

mahjonp: or change the panic message here.

pkg/control/control.go
```diff
@@ -488,18 +518,59 @@ func (c *Controller) queryTiDBClusterLogs(dur time.Duration, nodes []clusterType
...
+		switch n.Component {
+		case clusterTypes.TiDB, clusterTypes.TiKV, clusterTypes.PD:
+			wg.Add(1)
+
+			// containerName is added to be sure that we only fetch the logs of
+			// the container we want, not included sidecar containers.
+			go func(ns, podName, containerName string) {
+				defer wg.Done()
+				texts, err := c.lokiClient.FetchPodLogs(ns, podName, containerName,
+					"", nil, c.lastQueryTime, to, 20000, false)
```
yeya24, line 553: Is this enough? 20000 for 4 minutes. We just found one tikv produce 20w logs in 2 minutes when debug is opened @mahjonp

mahjonp: Maybe we should reduce the duration of each fetching logs? How about 20000/min?

yeya24:
Updated @mahjonp

PR #213, opened 2020-04-28
Add loki parameters to argo yaml files

yeya24:

This bug is fixed in the tidb-operator side, so we don't need to explicitly enable this anymore.

This PR also adds some missing part to argo yaml files


yeya24:
We still need to enable the `placement rule` in our code because of #2328. So remove that change

yeya24:
The latest commit fixes the concurrent map write bug, also closes all logs files at the end of test. PTAL @mahjonp

PR #214, opened 2020-04-29
update cdc server startup parameter

amyangfei:

Add some startup parameter in cdc

- set `log-level` to debug
- set `tz` to UTC

### Check List 

Tests 

 - Unit test
 - E2E test

### DIFF
--- a/pkg/test-infra/cdc/recommend.go
+++ b/pkg/test-infra/cdc/recommend.go
@@ -45,6 +45,8 @@ func newCDC(ns, name string) *CDC {
 			"app.kubernetes.io/component": "cdc",
 			"app.kubernetes.io/instance":  cdcName,
 		}
+		logLevel = fixture.Context.CDCConfig.LogLevel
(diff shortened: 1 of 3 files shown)


pkg/test-infra/cdc/recommend.go
```diff
@@ -91,7 +91,9 @@ func newCDC(ns, name string) *CDC {
 									"server",
 									fmt.Sprintf("--pd=%s", fmt.Sprintf("http://%s:2379", upstreamPDAddr)),
 									"--status-addr=0.0.0.0:8300",
+									"--log-level", "debug",
```
yeya24, line 94: What about making the log level as a configurable flag? You can add flags here https://github.com/pingcap/tipocket/blob/4f56639edf2b27712f87bb41a3d225236bef38e2/pkg/test-infra/fixture/fixture.go#L246

amyangfei:
PTAL @yeya24 thanks!

PR #222, opened 2020-05-09
elle: add graphviz

mahjonp:

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


siddontang:
can you show an example about how graphviz represents the result here?

mahjonp:
For example, we have a test TestPlotAnalysis.

https://github.com/pingcap/tipocket/blob/1cb42372c3572b9afd2b9b898113e3ef662ab023/pkg/elle/list_append/viz_test.go#L10-L26

We can see that the three topmost nodes build a cycle(wr -> realtime -> realtime path), rw means `Ti installs xi , Tj reads x i`, realtime means `Ti has committed before Tj starts`, that's a G1c-realtime anomaly.

PR #228, opened 2020-05-11
Add kafka cron and other minor fixes

yeya24:

Seems there is an indentation problem in the previous workflow yamls. But I am curious why Argo doesn't have a validation webhook for this...

Add cdc Kafka workflow


yeya24:
Also add a cluster check functionality to don't run tipocket case until the cluster is clean up for cron workflow

PR #245, opened 2020-05-20
support tikv encryption test

mahjonp:
Signed-off-by: mahjonp <junpeng.man@gmail.com>

### What is changed and how does it work?

This PR will inject a master key(hardcoded) into tikv pod and use specified tikv.toml to activate encryption. 

### Check List 

Tests 

 - Unit test
 - E2E test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has Go code change
 - Has CI related scripts change
 - Has Terraform scripts change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
NONE
```

PR #248, opened 2020-05-20
Unify all the flags between tipocket and argo

yeya24:

Fixes https://github.com/pingcap/tipocket/pull/245#discussion_r428067101

Unify the parameters between tipocket and argo.

Note that for the flag with `.` such as `cdc.version` or `abtest.general-log`, I rename `.` to `_` such as `cdc_version` or `abtest_general-log` because argo name doesn't allow `.`


mahjonp:
LGTM

PR #251, opened 2020-05-21
Fix sqllogic case parameter

hunterlxt:
Fix sqllogic case parameter

Signed-off-by: TXXT <hunterlxt@live.com>


hunterlxt:
@mahjonp  PTAL

PR #258, opened 2020-05-27
DM: use sync-diff-inspector to check data

csuzhangxc:

DM may need to compare shard merge tables later, the `compareData` we used currently is not enough.

use our sync-diff-inspector to check data for DM tasks.

### Check List 

Tests 

 - Unit test


csuzhangxc:
@mahjonp @yeya24 PTAL

cmd/dm-pocket/main.go
```diff
@@ -32,6 +34,11 @@ var (
 
 func main() {
 	flag.Parse()
+
+	go func() {
```
mahjonp, line 38: Tipocket opens the pprof by default so you don't need to open it again.

https://github.com/pingcap/tipocket/blob/35dea2cc300e17a19aebbba537d5e01a5e390cc8/pkg/test-infra/fixture/fixture.go#L277-L279

csuzhangxc: Oh, thanks a lot, removed in https://github.com/pingcap/tipocket/pull/258/commits/473df6662695c10bd3934b83cce7b83b65e6d02d.

BTW, mirror update in https://github.com/pingcap/tipocket/pull/258/commits/d5336e7e5d410f87a403111f285e329113d936bc.

PR #263, opened 2020-06-03
Implement rwregister

you06:

Implement elle's [rw register module](https://github.com/jepsen-io/elle/blob/master/src/elle/rw_register.clj).

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - None

```release-note
Implement elle's rw register module
```


mahjonp:
It's ready for review? @you06

you06:
> It's ready for review? @you06

Not ready, some tests still got failed.

you06:
@mahjonp it's ready for review.

pkg/elle/core/depend.go
```diff
@@ -32,6 +32,14 @@ const (
 	WRDepend DependType = "wr"
 	// RWDepend ...
 	RWDepend DependType = "rw"
+	// EXTKeyDepend ...
```
mapleFU, line 35: emm...Should these be anomalies?

pkg/elle/core/history.go
```diff
@@ -96,16 +116,47 @@ func (m Mop) IsEqual(n Mop) bool {
...
 	switch m.T {
 	case MopTypeRead:
-		key := m.M["key"].(string)
-		value := m.M["value"]
-		if value == nil {
-			return fmt.Sprintf("[:r %s nil]", key)
+		key, ok := m.M["key"]
+		if ok {
+			value := m.M["value"]
+			if value == nil {
```
mapleFU, line 122: (Maybe just `fmt.Sprintf("[:r %s %v]", key, value)` is ok)

pkg/elle/core/history.go
```diff
@@ -205,12 +295,12 @@ func (b SameKeyOpsByLength) Swap(i, j int) {
 }
 
 // AttachIndexIfNoExists add the index for history with it's number in array.
-func (h History) AttachIndexIfNoExists() {
-	if len(h) != 0 && h[0].Index.Present() {
+func (h *History) AttachIndexIfNoExists() {
```
mapleFU, line 298: Since `Append` is not used here, why we need pointer?

pkg/elle/rw_register/op.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"regexp"
+	"strconv"
+
+	"github.com/pingcap/tipocket/pkg/elle/core"
+)
+
+var (
+	opPattern = regexp.MustCompile(`([rw])([a-zA-Z])([0-9_]+)(.*)`)
```
mapleFU, line 11: so 1 means "w/r"m, 2 means "key", 3means "keynumber", can you give some string samples here?

pkg/elle/rw_register/op.go
```diff
@@ -0,0 +1,77 @@
...
+				mopKey: mopVal,
+			},
+		})
+	}
+
+	return op
+}
+
+// Pair ...
+func Pair(op core.Op) (core.Op, core.Op) {
```
mapleFU, line 66: How will this used? for testing?

pkg/elle/rw_register/rw_register.go
```diff
@@ -1 +1,902 @@
...
+				if vprt == nil {
+					panic("write value should not be nil")
+				}
+				dataMap[k] = *vprt
+			}
+		}
+		if mop.IsRead() {
+			for k, v := range mop.M {
+				vprt := v.(*int)
+				if vprt == nil {
```
mapleFU, line 58: I think there is a logic in consistence, if vprt is nil, but `dataMap[k]` os not nil, it should report

pkg/elle/core/depend.go
you06: A cycle would be an anomaly, the depend explains the steps in it only.

pkg/elle/rw_register/op.go
you06: `rx1` => `[:r :x 1]`
`rx_` => `[:r :x nil]`
`wx1` => `[:w :x 1]`

pkg/elle/rw_register/op.go
you06: Yes, the functions in op will used by testing code only now.

pkg/elle/rw_register/rw_register.go
you06: This should be report, but the nil read will fallback to `initMagicNumber`, so there should not be nil reads. Add a panic here.

pkg/elle/core/depend.go
mapleFU: `DependType` is for single depend, maybe anomaly should be other type.

pkg/elle/rw_register/rw_register.go
mapleFU: Where's the logic? how can I read `1919810`?

you06:
@mapleFU 
I delete the `ext-key-graph` function and `EXTKeyDepend` because actually we don't need that, the find function in `transactionGraph2VersionGraph` is efficient enough.

For the following depends, we will use it in `CycleExplainerResult`, elle itself also did this. I checked out the tests in list append, where it just used a constant string and convert it to depend type. They are not depend types and I use a `string` instead of `DependType`.

```golang
// G1cDepend ...
G1cDepend DependType = "G1c"
// G2Item ...
G2Item DependType = "G2-item"
// GSingle ...
GSingle DependType = "G-single"
```

pkg/elle/core/history.go
you06: It'll be `[:r x <nil>]` if so, but I want `[:r x nil]`

pkg/elle/rw_register/rw_register.go
you06: ikuikuhaire?

pkg/elle/rw_register/utils.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+		}
+		for k := range mop.M {
+			keys[k] = struct{}{}
+		}
+	}
+	return keys
+}
+
+func extReadKeys(op core.Op) map[string]int {
```
mahjonp, line 151: Maybe set the return type as map[string]*int is better?

mahjonp: Because though the signature`map[string]int`, we can't tell the situations of 
1. reading the initial value of a key or 
2. txn haven't observed that key.

pkg/elle/rw_register/rw_register.go
```diff
@@ -1 +1,902 @@
...
+
+func wfrVersionGraphs(history core.History) map[string]*core.DirectedGraph {
+	vgs := make(map[string]*core.DirectedGraph)
+	for _, op := range core.FilterOkHistory(history) {
+		var (
+			reads  = extReadKeys(op)
+			writes = extWriteKeys(op)
+		)
+		for k, r := range reads {
+			w, ok := writes[k]
```
mahjonp, line 435: See my comments on `extReadKeys`. We should check whether r is nill.

pkg/elle/rw_register/op_test.go
```diff
@@ -0,0 +1,133 @@
...
+	require.Equal(t, expect, op)
+
+	op = MustParseOp("wx1")
+	expect = core.Op{
+		Type: core.OpTypeOk,
+		Value: &[]core.Mop{
+			{
+				T: core.MopTypeWrite,
+				M: map[string]interface{}{
+					"x": IntPtr(1),
```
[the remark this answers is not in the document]
you06: I want read & write use the same types.

pkg/elle/rw_register/rw_register.go
```diff
@@ -1 +1,902 @@
...
+		}
+	}
+
+	return anomalies
+}
+
+type wrExplainResult struct {
+	Typ   core.DependType
+	Key   string
+	Value core.MopValueType
```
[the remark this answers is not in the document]
you06: IMO `mop-index` does not make sense, there has no difference between find op before and after explain.

pkg/elle/rw_register/utils.go
you06: That makes sense.

you06:
I'll try removing `initMagicNumber`.

pkg/elle/rw_register/op.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+var (
+	opPattern = regexp.MustCompile(`([rw])([a-zA-Z])([0-9_]+)(.*)`)
+)
+
+type Int struct {
+	IsNil bool
+	Val   int
+}
```
you06, line 17: I prefer this struct being used by other cases, maybe it can be moved to `core` module.

pkg/elle/rw_register/rw_register_test.go
```diff
@@ -0,0 +1,947 @@
...
+	expect12x.Link(core.Vertex{Value: NewInt(1)}, core.Vertex{Value: NewInt(2)}, core.Realtime)
+	expect12x.Link(core.Vertex{Value: NewInt(1)}, core.Vertex{Value: NewInt(3)}, core.Realtime)
+	expect12 := map[string]*core.DirectedGraph{
+		"x": expect12x,
+		"y": core.NewDirectedGraph(),
+	}
+	g12 := core.NewDirectedGraph()
+	g12.Link(core.Vertex{Value: history12[0]}, core.Vertex{Value: history12[1]}, core.Realtime)
+	g12.Link(core.Vertex{Value: history12[0]}, core.Vertex{Value: history12[2]}, core.Realtime)
+	g12.Link(core.Vertex{Value: history12[2]}, core.Vertex{Value: history12[3]}, core.Realtime)
```
you06, line 370: The graph of this case

```
        ┌T2(wx2)
T1(wx1)─┤
        └t3(wy2)─T4(wx3)
```

`transactionGraph2VersionGraphs` see through other keys so we can infer the version graph for `x` from it:

```
  ┌2
1─┤
  └3
```

pkg/elle/core/history.go
```diff
@@ -39,17 +39,20 @@ type MopType string
...
 const (
-	OpTypeInvoke OpType = "invoke"
-	OpTypeOk     OpType = "ok"
-	OpTypeFail   OpType = "fail"
-	OpTypeInfo   OpType = "info"
+	OpTypeInvoke  OpType = "invoke"
+	OpTypeOk      OpType = "ok"
+	OpTypeFail    OpType = "fail"
+	OpTypeInfo    OpType = "info"
+	OpTypeUnknown OpType = "unknown"
```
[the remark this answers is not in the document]
you06: I prefer leaving this constant here.

pkg/elle/rw_register/rw_register_test.go
you06: We still don't need `ext-key-graph`, the `find` closure function will be much easier.

mahjonp: OK, but memorize `find` can make it more efficient.

pkg/elle/core/graph.go
```diff
@@ -462,6 +485,81 @@ func (g *DirectedGraph) RenumberGraph() (*DirectedGraph, func(interface{}) inter
...
+	case string:
+		return val
+	case int:
+		return strconv.Itoa(val)
+	case Op:
+		return val.String()
+	case KV:
+		return val.String()
+	default:
+		return CallMethod(val, "String").(string)
```
mahjonp, line 499: How about type assertion to `fmt.Stringer`?

pkg/elle/core/history.go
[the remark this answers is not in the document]
mahjonp: What's the purpose of retaining it? There is no event type of 'unknown' actually.

pkg/elle/core/history.go
```diff
@@ -392,3 +482,13 @@ func (h History) GetKeys(t MopType) []string {
 
 	return keys
 }
+
+// KV struct
+type KV struct {
```
mahjonp, line 487: KVEntity maybe better.

pkg/elle/rw_register/op.go
mahjonp: What's the difference between IntOptional and Int? Can we use IntOptional instead of creating another similar type?

pkg/elle/rw_register/rw_register.go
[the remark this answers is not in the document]
mahjonp: aMopIndex and bMopIndex are needed when using Graphviz to render the dependency graph

mahjonp:
Add some new comments. PTAL @you06

PR #267, opened 2020-06-12
dm: decrease check duration; add pod kill nemesis

csuzhangxc:

- TiDB may not catchup MySQL quickly
- need to add nemesis for DM

- decrease check duration from 1m to 10s
- add pod kill nemesis for DM

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

```release-note
NONE
```


csuzhangxc:
@mahjonp PTAL

PR #293, opened 2020-07-20
Update argo with Matrix's config

oraluben:
### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


oraluben:
Current: Argo verified (but slow and an error occurs after long-time execution)
Todo: Do not modify templates, create new workflows to not break current ones.

Dockerfile
```diff
@@ -6,6 +6,7 @@ ENV GO111MODULE=on
 WORKDIR /src
 COPY . .
 
+RUN make clean
```
mahjonp, line 9: I think `make clean` is not need here.

oraluben: I added this to not copy the local bin when testing with a single binary.

PR #318, opened 2020-10-30
elle: mark invalid connection in commit as unknown state

you06:

There are some invalid connection errors in commit stage, but actually we don't know whether this statement is executed successful or not. Currently we mark these transactions as failed, this may wrongly result to G1a phenomenon if the transactions are successfully committed. 

If the commit error is invalid connection error, we mark this transaction to unknown state.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - This change may miss some real G1a phenomenon.

Related changes

 - None

### DIFF
--- a/pkg/cluster/manager/types/resource.go
+++ b/pkg/cluster/manager/types/resource.go
@@ -2,9 +2,10 @@ package types
 
 import (
 	"fmt"
-	"github.com/jinzhu/gorm"
 	"strconv"
 	"strings"
+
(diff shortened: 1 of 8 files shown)


you06:
@mahjonp I've checked that G1a will only find read from failed operations and also unknown state operations won't be compose into the dependency graph, PTAL.

mahjonp (approved):
lgtm

PR #320, opened 2020-11-05
add argo files for bank2-compaction

cosven:
Run bank2 testcase with `compaction guard` and `gc in compaction filter` turned on. The testcase is supposed to run every 06:00 and last for 3 hours.

cc @yiwu-arbug @hicqu


config/tikv/compaction.toml
```diff
@@ -0,0 +1,23 @@
...
+force-consistency-checks = true
+
+[rocksdb.defaultcf]
+enable-compaction-guard = true
+
+[rocksdb.writecf]
+enable-compaction-guard = true
+
+[rocksdb.lockcf]
+enable-compaction-guard = true
```
mahjonp, line 23: A newline is needed here.

mahjonp (approved):
lgtm

config/tikv/compaction.toml
yiwu-arbug, line 23: we don't plan to enable compaction guard for lockcf.

config/tikv/compaction.toml
```diff
@@ -0,0 +1,23 @@
...
+consistency-check-method = "mvcc"
+
+[raftstore]
+consistency-check-interval = "180s"
+
+[rocksdb]
+force-consistency-checks = true
+
+[rocksdb.defaultcf]
+enable-compaction-guard = true
```
yiwu-arbug, line 17: let's also set `target-file-size-base = 128M`. This config will default to 128M when we default on compaction guard, but not right now.

argo/cron/bank2-compaction.yaml
```diff
@@ -0,0 +1,50 @@
+# maintainer: cosven
+
+metadata:
+  name: tipocket-bank2-compaction
```
yiwu-arbug, line 4:
```suggestion
  name: tipocket-bank2-compaction-opt
```
this is testing some compaction related optimization, not compaction itself. Also change the name of bank2-compaction.yaml.

yiwu-arbug (approved):
LGTM

PR #327, opened 2020-12-02
Add async commit and 1PC parameters for some cases

ekexium:

Since pingcap/tidb#21365, async commit needs to be enabled by setting a system variable. So we need to change some cases.

Add `async-commit` and `one-pc` parameters to the `bank2` and `tpcc` cases, and set corresponding system variables in the `SetUp`.

### Check List 

Tests 

 - Manual test: manually test if these cases successfully set the system variables

Code changes

 - Has Go code change

```release-note
NONE
```


ekexium:
@mahjonp PTAL

mahjonp (approved):
lgtm

PR #330, opened 2020-12-04
Add support of base64://CONTENT in -tidb-config, -tikv-config and -pd-config

ofey404:

https://github.com/pingcap/tipocket/issues/324

`-tikv-config`, `-tidb-config`, `-pd-config` support base64 encoded configuration.

Example:
```bash
bin/bank2 ...{other flags}... -tikv-config="base64://bG9nLWxldmVsID0gIndhcm5pbmciCg=="
```

Generate base64 string
```bash
$ cat tikv-test-config.toml ; base64 tikv-test-config.toml 
log-level = "warning"
bG9nLWxldmVsID0gIndhcm5pbmciCg==
```

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Related changes

```release-note
add base64 support of -tikv-config, -tidb-config, -pd-config
eg:
    -tikv-config="base64://bG9nLWxldmVsID0gIndhcm5pbmciCg=="
```


pkg/test-infra/tidb/operation.go
```diff
@@ -459,6 +468,37 @@ func getTidbDiscoveryService(tc *v1alpha1.TidbCluster) *corev1.Service {
 	}
 }
 
+func parseConfig(config string) (string, error) {
```
cosven, line 471: How about checking scheme first and then fallback to old behaviour?
I think we need not to check if the config file exists here. It is checked in `readFileAsString`.
Think about the case when the config value is empty.

pkg/test-infra/tidb/operation.go
```diff
@@ -459,6 +468,37 @@ func getTidbDiscoveryService(tc *v1alpha1.TidbCluster) *corev1.Service {
...
+		}
+		var configData string = ""
+		switch u.Scheme {
+		case "base64":
+			// When parse url like "base64://BASE64CONTENT", BASE64CONTENT would be parsed in url.Host
+			configData, err = readBase64AsString(u.Host)
+			if err != nil {
+				return "", err
+			}
+			fmt.Print(configData)
```
cosven, line 495: remove the debug statement

cosven (commented):
rest LGTM

pkg/test-infra/tidb/operation.go
ofey404: You are right.

pkg/test-infra/tidb/operation.go
ofey404: removed.

pkg/test-infra/tidb/operation.go
```diff
@@ -459,6 +467,32 @@ func getTidbDiscoveryService(tc *v1alpha1.TidbCluster) *corev1.Service {
...
+	}
+	switch u.Scheme {
+	case "base64":
+		// An workaround: When parse url like "base64://BASE64CONTENT", BASE64CONTENT would be parsed in url.Host
+		configData, err = readBase64AsString(u.Host)
+		if err != nil {
+			return "", err
+		}
+	default:
+		return "", fmt.Errorf("not a supported scheme: %s\nsupported schemes like base64://BASE64CONTENT", config)
```
cosven, line 490: Generally, we do not insert `newline` in error message.

cosven (approved):

pkg/test-infra/tidb/operation.go
mahjonp: I agree with you.

mahjonp (approved):
LGTM 💯

pkg/test-infra/tidb/operation.go
ofey404: Have removed it.

PR #336, opened 2020-12-17
Support run preparation sqls for TiDB clusters

sticnarf:

Sometimes we need to set some global variables or do some other preparations to the TiDB cluster. With this PR, we can pass these preparation SQLs through arguments.

```release-note
NONE
```

### DIFF
--- a/pkg/core/db.go
+++ b/pkg/core/db.go
@@ -26,13 +26,21 @@ type NoopDB struct {
 
 // SetUp initializes the database.
 func (NoopDB) SetUp(ctx context.Context, nodes []cluster.Node, node cluster.Node) error {
-	prepareSQL := fixture.Context.TiDBClusterConfig.PrepareSQL
-	if len(prepareSQL) > 0 && node.Component == cluster.TiDB {
+	isFirstTiDB := false
+	for _, n := range nodes {
+		if n.Component == cluster.TiDB {
+			isFirstTiDB = n == node
+			break
+		}
+	}
+
+	// Run prepare SQLs
+	if isFirstTiDB {
 		db, err := sql.Open("mysql", fmt.Sprintf("root@tcp(%s:%d)/test?multiStatements=true", node.IP, node.Port))
 		if err != nil {
 			return err
 		}
-		if _, err = db.Exec(prepareSQL); err != nil {
+		if _, err = db.Exec(fixture.Context.TiDBClusterConfig.PrepareSQL); err != nil {
 			return err
 		}
 		if err = db.Close(); err != nil {
@@ -40,8 +48,6 @@ func (NoopDB) SetUp(ctx context.Context, nodes []cluster.Node, node cluster.Node
 		}
 		// Wait until global variables take effect
 		time.Sleep(2100 * time.Millisecond)
-		// Only need to run once
-		fixture.Context.TiDBClusterConfig.PrepareSQL = ""
 	}
 	return nil
 }
--- a/pkg/core/db.go
+++ b/pkg/core/db.go
@@ -2,7 +2,10 @@ package core
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
+	"github.com/pingcap/tipocket/pkg/test-infra/fixture"
+	"time"
 
 	"github.com/pingcap/tipocket/pkg/cluster"
 )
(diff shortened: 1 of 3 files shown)


pkg/test-infra/tidb/operation.go
```diff
@@ -189,6 +190,9 @@ func (o *Ops) Apply() error {
 	if err := o.waitTiDBReady(tc, fixture.Context.WaitClusterReadyDuration); err != nil {
 		return err
 	}
+	if err := o.runPreparationSQL(); err != nil {
```
mahjonp, line 193: I prefer to run the preparation sql on `func (c *Controller) setUpDB()`:

```
// SetUp initializes the database.
func (NoopDB) SetUp(ctx context.Context, nodes []cluster.Node, node cluster.Node) error {
        ...
	return nil
}
```

pkg/core/db.go
```diff
@@ -23,6 +26,23 @@ type NoopDB struct {
 
 // SetUp initializes the database.
 func (NoopDB) SetUp(ctx context.Context, nodes []cluster.Node, node cluster.Node) error {
+	prepareSQL := fixture.Context.TiDBClusterConfig.PrepareSQL
+	if len(prepareSQL) > 0 && node.Component == cluster.TiDB {
+		db, err := sql.Open("mysql", fmt.Sprintf("root@tcp(%s:%d)/test", node.IP, node.Port))
```
mahjonp, line 31: I suggest adding `multiStatements=true` To support multiple statements.

```suggestion
		db, err := sql.Open("mysql", fmt.Sprintf("root@tcp(%s:%d)/test?multiStatements=true", node.IP, node.Port))
```

pkg/core/db.go
```diff
@@ -2,7 +2,10 @@ package core
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
+	"github.com/pingcap/tipocket/pkg/test-infra/fixture"
```
mahjonp, line 7: reorg imports

mahjonp (commented):
Rest LGTM. @cosven PTAL

pkg/core/db.go
```diff
@@ -23,6 +26,23 @@ type NoopDB struct {
...
+		if _, err = db.Exec(prepareSQL); err != nil {
+			return err
+		}
+		if err = db.Close(); err != nil {
+			return err
+		}
+		// Wait until global variables take effect
+		time.Sleep(2100 * time.Millisecond)
+		// Only need to run once
+		fixture.Context.TiDBClusterConfig.PrepareSQL = ""
```
cosven, line 44: It seems that it may run for multiple times

```golang
func (c *Controller) syncNodeExec(f func(i int)) {
        var wg sync.WaitGroup
        n := len(c.cfg.Nodes)
        wg.Add(n)
        for i := 0; i < n; i++ {
                go func(i int) {
                        defer wg.Done()
                        f(i)
                }(i)
        }
        wg.Wait()
}

func (c *Controller) setUpDB() {
        log.Infof("begin to set up database")
        c.syncNodeExec(func(i int) {
                log.Infof("begin to set up database on %s", c.cfg.Nodes[i])
                db := core.GetDB(c.cfg.DB)
                err := db.SetUp(c.ctx, c.cfg.Nodes, c.cfg.Nodes[i])
                if err != nil {
                        log.Fatalf("setup db %s at node %s failed %v", c.cfg.DB, c.cfg.Nodes[i], err)
                }
        })
}
```

pkg/core/db.go
```diff
@@ -23,6 +26,23 @@ type NoopDB struct {
...
+			return err
+		}
+		if _, err = db.Exec(prepareSQL); err != nil {
+			return err
+		}
+		if err = db.Close(); err != nil {
+			return err
+		}
+		// Wait until global variables take effect
+		time.Sleep(2100 * time.Millisecond)
```
cosven, line 42: How about waiting in test case logic?

pkg/core/db.go
sticnarf: Yes. So we clear this config here to avoid running it multiple times.

pkg/core/db.go
sticnarf: It's only needed when global variables are modified. I think it may be not related to the test case itself.

mahjonp (approved):
lgtm

pkg/core/db.go
cosven: Emmm, `db.SetUp` runs concurrently. I think it may still run for multiple times even we clear the config here.

sticnarf: Ah, I see. It is a race as well.

cosven (approved):

PR #349, opened 2021-01-27
feat: support log volume binding

mahjonp:

Since the TiDB-Operator has supported mounting PV into containers, this PR updates recommend TiDB ops configuration with a log PV binding with TiDB container and save all main component log into PVs.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/test-infra/tidb/template.go
+++ b/pkg/test-infra/tidb/template.go
@@ -74,7 +74,8 @@ then
 fi
 echo "start tidb-server ..."
 echo "/tidb-server ${ARGS}"
-/tidb-server ${ARGS} 2>&1 | tee /var/log/tidblog/tidb.log
+/tidb-server ${ARGS} 2>&1 | tee -a /var/log/tidblog/tidb.log
+#                               ^^: append mode
 `))
 
 // StartScriptModel ...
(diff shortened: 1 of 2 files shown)


pkg/test-infra/tidb/template.go
```diff
@@ -74,7 +74,8 @@ then
 fi
 echo "start tidb-server ..."
 echo "/tidb-server ${ARGS}"
-exec /tidb-server ${ARGS}
+/tidb-server ${ARGS} 2>&1 | tee -a /var/log/tidblog/tidb.log
+#                               ^^: append mode
```
cosven, line 78: How about set log-file to `/var/log/tidblog/tidb.log` in config file?

Pros: 
* It will auto rotate log
* configurable
* daily run case are controlled by admin, we can ensure that these cases use correct config

mahjonp: That's troubling to me, I need such a tool to help me:
1. run the tikv-server in the background
2. log tail tikv.log to stdout
3. exit with the exit-code when tikv-server exits

mahjonp: resolve in 448bc24

mahjonp:
Resolve the log store and avoid loss issue that has recorded in #321

cosven:
LGTM
cc @ofey404

cosven (approved):

PR #357, opened 2021-02-02
chore: uses the `hostname.subdomain.namespace.svc` format node ip

mahjonp:

This PR uses the `hostname.subdomain.namespace.svc` format node ip

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -59,7 +59,7 @@ type Node struct {
 
 // String ...
 func (node Node) String() string {
-	return fmt.Sprintf("node[ns=%s,ip=%s:%d]", node.PodName, node.IP, node.Port)
+	return fmt.Sprintf("node[ns=%s,ip=%s:%d]", node.Namespace, node.IP, node.Port)
 }
 
 // ClientNode is TiDB's exposed endpoint, can be a nodeport, or downgrade cluster ip
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -59,7 +59,7 @@ type Node struct {
 
 // String ...
 func (node Node) String() string {
-	return fmt.Sprintf("%s %s:%d", node.PodName, node.IP, node.Port)
+	return fmt.Sprintf("node[ns=%s,ip=%s:%d]", node.PodName, node.IP, node.Port)
 }
 
 // ClientNode is TiDB's exposed endpoint, can be a nodeport, or downgrade cluster ip
@@ -78,7 +78,7 @@ func (clientNode ClientNode) Address() string {
 
 // String ...
 func (clientNode ClientNode) String() string {
-	return fmt.Sprintf("%s %s:%d", clientNode.ClusterName, clientNode.IP, clientNode.Port)
+	return fmt.Sprintf("client_node[ns=%s,ip=%s:%d]", clientNode.Namespace, clientNode.IP, clientNode.Port)
 }
 
 // Cluster interface
--- a/pkg/test-infra/tidb/operation.go
+++ b/pkg/test-infra/tidb/operation.go
@@ -613,7 +613,7 @@ func getTiDBServicePort(svc *corev1.Service) int32 {
 			return port.Port
 		}
 	}
-	return 0
+	panic("couldn't find the tidb exposed port")
 }
 
 // GetTiDBConfig is used for Matrix-related setups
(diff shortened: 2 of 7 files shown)


pkg/cluster/cluster.go
```diff
@@ -59,7 +59,7 @@ type Node struct {
 
 // String ...
 func (node Node) String() string {
-	return fmt.Sprintf("%s %s:%d", node.PodName, node.IP, node.Port)
+	return fmt.Sprintf("node[ns=%s,ip=%s:%d]", node.PodName, node.IP, node.Port)
```
cosven, line 62: ns=podname?

pkg/test-infra/tidb/operation.go
```diff
@@ -108,23 +111,36 @@ func (o *Ops) GetNodes() ([]cluster.Node, error) {
...
+		}
+		clientNodes = append(clientNodes, cluster.ClientNode{
+			Namespace:   o.ns,
+			ClusterName: o.name,
+			IP:          svc.Spec.ClusterIP,
+			Port:        getTiDBServicePort(svc),
+		})
+	} else {
+		// If case isn't running on k8s pod, uses nodeIP:tidb_port as clientNode for conveniently debug on local
+		ips, err := util.GetNodeIPs(o.cli, o.ns, map[string]string{"app.kubernetes.io/instance": o.name})
```
cosven, line 127: It seems that this will not work on our new K8s since the name looks like `idc-5-46`.

pkg/test-infra/tidb/operation.go
```diff
@@ -108,23 +111,36 @@ func (o *Ops) GetNodes() ([]cluster.Node, error) {
...
+	if util.IsInK8sPodEnvironment() {
+		svc, err := o.getTiDBServiceByClusterName(o.ns, o.name)
+		if err != nil {
+			return nil, err
+		}
+		clientNodes = append(clientNodes, cluster.ClientNode{
+			Namespace:   o.ns,
+			ClusterName: o.name,
+			IP:          svc.Spec.ClusterIP,
+			Port:        getTiDBServicePort(svc),
```
cosven, line 123: Maybe we shoud raise an error If the `getTiDBServicePort` returns 0?

pkg/cluster/cluster.go
mahjonp: Thanks and resolved on 504f4fdfe0c72ac9fc445a264a5d3469c991b9d9.

pkg/test-infra/tidb/operation.go
mahjonp: I rename it to `GetNodeIPsFromPod(cli client.Client, namespace string, podLabels map[string]string)`, the 3rd parameter is the pod labels.

pkg/test-infra/tidb/operation.go
mahjonp: Resolved on 7c88cd78099cc909c51233ae5538d27172fcac22. For convenient, panic if not expect.

pkg/test-infra/tidb/operation.go
```diff
@@ -108,23 +111,36 @@ func (o *Ops) GetNodes() ([]cluster.Node, error) {
...
+		if err != nil {
+			return nil, err
+		} else if len(ips) == 0 {
+			return nil, errors.New("k8s node not found")
+		}
+		svc, err := o.getTiDBServiceByClusterName(o.ns, o.name)
+		if err != nil {
+			return clientNodes, err
+		}
+		clientNodes = append(clientNodes, cluster.ClientNode{
```
ofey404, line 137: Maybe it's off-topic, but it seems that `GetClientNodes()` will only return one `ClientNode` in an array. Is this for backward compatibility?

ofey404 (commented):
Look good to me.

pkg/test-infra/tidb/operation.go
mahjonp: The concept `clientNode` is borrowed from Jepsen but there are a few differences against Jepsen.

In Jepsen, each client connects to different tidb-servers(deployed on docker-swarm), but because of the existence of service(as a load balancer) on K8s, the client can connect to tidb-servers through the service, so we simplify it to only one ClientNode on TiPocket.

cosven (approved):

PR #358, opened 2021-02-03
*: uses Jsonnet to generate argo workflow

### LINKED ISSUE #208: manage argo workflows or templates using config management tools
## Feature Request

It would be nice if we can manage the argo workflows and argo templates using some config management tools like helm, jsonnet or kustomize.

It is really a pain to manage all of these manually.

### LINKED ISSUE #355: randomize generated namespace name
## Feature Request

**Is your feature request related to a problem? Please describe:**

When apply workflow to argo, the workflow will stuck if there already is a tidb cluster in given namespace.
It can be triggered by applying same workflow twice.

**Describe the feature you'd like:**

Add a ramdom suffix to generated namespace name, so that workflows can run in different namespaces.

mahjonp:

Handwriting argo workflow is very troublesome.

close #355 close #208  and also #321 case writing part:

```
编写完用例后，需要编写相应的 yaml 文件
yaml 文件大部分内容是 重复 的，而且文件行数比较多，编写、review、维护 都很心累
```

This PR uses Jsonnet to make life easy.

For example, If I have written a test case named `list-append` which has three parameters: tablecount, read_lock, txn_mode.

To let it runnable, I need to fill on the metadata about list-append case on [lib/case.jsonnet](./lib/case.jsonnet) just like below:

```jsonnet
list_append(args={ tablecount: '7', read_lock: '"FOR UPDATE"', txn_mode: 'pessimistic' })::
  [
    '/bin/append',
    '-table-count=%s' % args.tablecount,
    '-read-lock=%s' % args.read_lock,
    '-txn-mode=%s' % args.txn_mode,
  ],
```

Looks simple right? Jsonnet is a simple extension of JSON, on above I define a function named list_append which has three arguments, and returns an array.

After fill on the testcase metadata, I should write a workflow file on [workflow/list-append.jsonnet](./workflow/list-append.jsonnet), it's content:

```jsonnet
{
  _config+:: {
    case_name: 'list_append',
    image_name: 'hub.pingcap.net/qa/tipocket',
    args+: {
      // k8s configurations
      'storage-class': 'local-storage',
      // client configurations
      client: 5,
      'request-count': 100000,
      round: 10,
    },
    command: { tablecount: '7', read_lock: '"FOR UPDATE"', txn_mode: 'pessimistic' },
  },
}
```

Explain a little here, `_config+` means override some key-value defined on [lib/config.jsonnet](./lib/config.jsonnet).

On the `args+` part, I set `storage-class`, `client` and `round` etc parameters declared on the `fixture.go` file which is the common flag all over test cases.

Then, the `command` field is set to invoke list_append case with 7 tables, use `FOR UPDATE` as the read_lock.

Now I could run `make build_workflow` to generate the workflow file.

```bash
$ make build_workflow
find run -name "*.jsonnet" | xargs -I{} jsonnetfmt -i {}
rm -rf run/manifest/*
find run/workflow -name "*.jsonnet" -type f -exec basename {} \;  | xargs -I% sh -c 'jsonnet run/workflow/% -J run/lib | yq eval -P - > run/manifest/%.yaml'
```

Lucky, everything goes well. ☺

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cosven (approved):
LGTM

yeya24:
This is great! Perfect work.

run/lib/argo/cron_workflow.jsonnet
```diff
@@ -0,0 +1,91 @@
+{
```
yeya24, line 1: Usually, libraries use `.libsonnet` as file extension name.

mahjonp: Thanks for your comment. Got it. @yeya24

PR #377, opened 2021-03-02
tests/vbank: add a flag to set conn params

zyguan:
Signed-off-by: zyguan <zhongyangguan@gmail.com>

Cannot specify connection parameters currently.

### Check List 

Add a flag to allow us set connection parameters so that we can enable some new features (like async-commit, clustered index) by system variables, (eg. `-conn_params='tidb_enable_async_commit=1&tidb_enable_1pc=1'`)

Tests 

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - None

```release-note
NONE
```


mahjonp (approved):
lgtm

sticnarf (approved):
LGTM. I just know this way. Then we don't need `prepareSql` to enable async commit/1PC.

mahjonp:
> LGTM. I just know this way. Then we don't need `prepareSql` to enable async commit/1PC.

I like this way too.

PR #378, opened 2021-03-03
add a compaction filter long term case

hicqu:
Signed-off-by: qupeng <qupeng@pingcap.com>

Add a long term case for GC in Compaction Filter.

The case has such features:
* Has both long records and short records, which covers different code path for GC in compaction filter.
* 3 SQLs are contianed in its workload: update/delete/insert.
* 4 actions on TiKV side: MVCC_PUT(record), MVCC_PUT(index), MVCC_DELETE(record), MVCC_DELETE(index). 
* A simple transaction consistency checker during running, like bank2, but no transaction log tracing.
* `admin check table` for every round.

### TODO list
* [ ] Support to automatic run it on tipocket.

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

```release-note
NONE
```

### DIFF
--- a/tests/gc-in-compaction-filter/README.md
+++ b/tests/gc-in-compaction-filter/README.md
@@ -1,0 +1,14 @@
+### Prepare Data
+
+```
+$ sysbench --mysql-host=<host> --mysql-port=<port> --mysql-user=root --tables=128 --table-size=1000000 --threads=32 --time=600 common.lua prepare
+```
+
+### Run the Test Case
+
+```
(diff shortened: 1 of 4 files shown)


mahjonp:
> * [ ]  Support to automatic run it on tipocket.

This test case couldn't run on tipocket now since the dependencies are lack in the docker image, I plan to let each test case owns a Dockerfile so that dependencies and manifests(Lua, Sysbench, etc) can be contained on the split images. 

Until then, this issue is resolved.

PR #380, opened 2021-03-04
add stale read test

NingLin-P:
### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


testcase/stale-read/client.go
```diff
@@ -0,0 +1,319 @@
...
+
+// NextRequest
+func (s *staleRead) NextRequest() interface{} {
+	panic("implement me")
+}
+
+// DumpState
+func (s *staleRead) DumpState(ctx context.Context) (interface{}, error) {
+	panic("implement me")
+}
```
mahjonp, line 319: All these methods aren't necessary, you can delete them.

mahjonp (commented):
Rest LGTM, I just take a brief look

testcase/stale-read/client.go
```diff
@@ -0,0 +1,306 @@
...
+			return
+		default:
+			time.Sleep(s.RequestInterval)
+			count++
+		}
+	}
+}
+
+func (s *staleRead) staleRead(ctx context.Context, conn *sql.Conn, t time.Time, ids []int) (map[int]string, uint64, error) {
+	txn, err := conn.BeginTx(ctx, &sql.TxOptions{})
```
Yisaer, line 194: `BeginTx` here will directly send `begin` to sql server. I think we should use `ExecuteContext` to begin a stale transaction by `conn.BeginTx(ctx, "START TRANSACTION READ ONLY AS OF TIMESTAMP...") `

Yisaer (commented):
REST lgtm

testcase/stale-read/client.go
NingLin-P: `conn.BeginTx(ctx, "START TRANSACTION READ ONLY AS OF TIMESTAMP...")` not match the function signature `func (*sql.Conn).BeginTx(ctx context.Context, opts *sql.TxOptions)`, I use the `SET TRANSACTION READ ONLY AS OF TIMESTAMP` instead, PTAL

Yisaer (approved):

PR #388, opened 2021-03-17
*: allow to customize cluster-name by flag

zyguan:
Signed-off-by: zyguan <zhongyangguan@gmail.com>

There are too many combinations even for running a single case binary, eg. test args, tidb versions. Currently we create a namespace for each test run, which may lead to too many namespaces.

This PR allow user to customize cluster-name by flag, so that we can run cases in same namespace.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - None

```release-note
NONE
```


mahjonp (approved):
lgtm

PR #390, opened 2021-03-18
skip creating namespace when the namespace is existed

Thearas:
Signed-off-by: Thearas <thearas850@gmail.com>

A test may fail at `TestClient.CreateNamespace` when using a low-privileged K8S user which lacks the permission to create a namespace, even if the namespace it needs already exists.

Skip creating or updating namespace if the namespace is already existed.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - None

```release-note
NONE
```


mahjonp (approved):
lgtm

PR #394, opened 2021-03-29
cross-region: support to specify cluster name prefix

Thearas:
Signed-off-by: Thearas <thearas850@gmail.com>

We may run multiple cross-region tests concurrently in one namespace, but it's impossible for now because all tidb clusters created by cross-region has the same name.

Support `cluster-name` flag. It will be a prefix of every cluster(a.k.a. `dcLocation`) name.

See also:
The PR of adding `cluster-name` flag: https://github.com/pingcap/tipocket/pull/388.

### Check List

Tests

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


testcase/cross-region/tso.go
```diff
@@ -70,17 +70,17 @@ func (c *crossRegionClient) testTSO(ctx context.Context) error {
...
 	}
 	log.Info("new allocators ready")
 	// after transfer allocator and leader, we can tolerate tso failed once for each dc.
 	c.requestTSOAfterTransfer(ctx, "global")
-	c.requestTSOAfterTransfer(ctx, "dc-1")
-	c.requestTSOAfterTransfer(ctx, "dc-2")
-	c.requestTSOAfterTransfer(ctx, "dc-3")
+	c.requestTSOAfterTransfer(ctx, DCLocations[0])
+	c.requestTSOAfterTransfer(ctx, DCLocations[1])
+	c.requestTSOAfterTransfer(ctx, DCLocations[2])
```
JmPotato, line 83: I think using a loop here to iter is better.

JmPotato (commented):
The Rest LGTM.

testcase/cross-region/tso.go
```diff
@@ -70,17 +70,17 @@ func (c *crossRegionClient) testTSO(ctx context.Context) error {
 		return err
 	}
 	log.Info("new leader ready")
-	for i := 1; i <= dcLocationNumber; i++ {
-		if err := c.transferPDAllocator(fmt.Sprintf("dc-%d", i)); err != nil {
+	for i := 0; i < dcLocationNumber; i++ {
```
zyguan, line 73: `len(DCLocations)` may be better.

testcase/cross-region/tso.go
Thearas: Ack.

testcase/cross-region/cross-region.go
```diff
@@ -34,6 +34,8 @@ add placement policy
 	role=leader
 	replicas=1`
 
+var DCLocations = []string{}
```
zyguan, line 37: If the number of DCs must be 3, then the type of `DCLocations` can be `[3]string`.

testcase/cross-region/tso.go
zyguan: Use for loop?

testcase/cross-region/cross-region.go
Thearas: I think the number of dc(s) may change in future, use slice here is more flexible.

testcase/cross-region/tso.go
Thearas: ok

testcase/cross-region/cross-region.go
```diff
@@ -198,7 +195,7 @@ func (c *crossRegionClient) setupDB() error {
 		return err
 	}
 	for i := 1; i <= 3; i++ {
-		stmt := fmt.Sprintf(stmtPolicy, fmt.Sprintf("p%v", i), fmt.Sprintf("dc-%v", i))
+		stmt := fmt.Sprintf(stmtPolicy, fmt.Sprintf("p%v", i), DCLocations[i-1])
```
JmPotato, line 198: Why not just start `i` form 0? `i-1` here seems not necessary.

Thearas: Because the partition name are formatting by `fmt.Sprintf("p%v", i)`.

JmPotato (approved):

zyguan (approved):
LGTM

Thearas:
PTAL @mahjonp .

PR #408, opened 2021-04-22
Use different test logic for different TSOs

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

Use different test logic for different TSOs.

* For the Global TSO test, we will request Local TSO twice before and after a Global TSO request and compare it with them.
* For the Local TSO test, we will compare it with the previous Global and Local TSO.

### Check List 

Tests 

 - Unit test
 - E2E test

```release-note
NONE
```


Yisaer (approved):

Thearas (approved):
LGTM

PR #411, opened 2021-05-26
Add minimal privileges test for backup using dynamic privs

morgo:

This does a minimal privilege test for backup using the `BACKUP_ADMIN` privilege.

Previously, `BACKUP` required the `SUPER` privilege, but this is not ideal from a principal of least privilege perspective. It is preferrable to instead have fine-grained task oriented privileges so that a user that performs the backup can not necessarily read the data.

I know it looks like it is a bit of a trivial test, but this validates that the minimal privileges do infact work for creating backups, and maybe eventually for restoring backups (see: https://github.com/pingcap/tidb/issues/24912 ).

It is my first experience writing TiPocket tests, and so I am happy to hear your suggestions on how this can be improved. I just followed the demo example scaffolding and mostly figured things out.

No specific change.

### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

Side effects

 - None
 
Related changes

 - None
 
There is no imcompatibility or change - if users want to keep allocating `SUPER` privilege to backup operators, they can still do that. However, we should encourage principal of least privilege to our users for security best practice. So from that perspective, there is a user-facing change once dynamic privileges are said to be stable.


morgo:
It looks like there is an incompatibility in go versions:
```
Error: ./client.go:81:19: undefined: os.MkdirTemp
note: module requires Go 1.15
```

mahjonp:
> It looks like there is an incompatibility in go versions:
> 
> ```
> Error: ./client.go:81:19: undefined: os.MkdirTemp
> note: module requires Go 1.15
> ```

The github-actions workflow uses [go 1.13](https://github.com/pingcap/tipocket/blob/master/.github/workflows/build.yaml#L10), update to [a higher version](https://github.com/actions/go-versions/blob/main/versions-manifest.json#L3) could solve it.

morgo:
> > It looks like there is an incompatibility in go versions:
> > ```
> > Error: ./client.go:81:19: undefined: os.MkdirTemp
> > note: module requires Go 1.15
> > ```
> 
> Hi @morgo, the github-actions workflow uses [go 1.13](https://github.com/pingcap/tipocket/blob/master/.github/workflows/build.yaml#L10), update to [a higher version](https://github.com/actions/go-versions/blob/main/versions-manifest.json#L3) could solve it.

Can I just upgrade it, or is this something that needs to be discussed/tested by others first? I could use ioutil.TempDir, but that is deprecated in 1.16, so it is better of course if I can upgrade it.

mahjonp:
> > > It looks like there is an incompatibility in go versions:
> > > ```
> > > Error: ./client.go:81:19: undefined: os.MkdirTemp
> > > note: module requires Go 1.15
> > > ```
> > 
> > 
> > Hi @morgo, the github-actions workflow uses [go 1.13](https://github.com/pingcap/tipocket/blob/master/.github/workflows/build.yaml#L10), update to [a higher version](https://github.com/actions/go-versions/blob/main/versions-manifest.json#L3) could solve it.
> 
> Can I just upgrade it, or is this something that needs to be discussed/tested by others first? I could use ioutil.TempDir, but that is deprecated in 1.16, so it is better of course if I can upgrade it.

There are some work items that should perform:

- [ ] Change the go version on every case
- [ ] Update the scaffold tool
- [ ] Update the baseimage on Dockerfile
- [ ] Update CI workflow scripts

If you would like, how about using ioutil.TempDir on this PR and I will give a PR about updating the go version ASAP.

morgo:
> If you would like, how about using ioutil.TempDir on this PR and I will give a PR about updating the go version ASAP.

That sounds good to me. I will make it work with 1.13 with `ioutil`.

mahjonp:
Hi, @morgo I'm curious why choose TiPocket to add this test case? IMO, this is a functional test case and https://github.com/pingcap/automated-tests/tree/master/ticases seems more suit for writing this kind of case.

morgo:
> Hi, @morgo I'm curious why choose TiPocket to add this test case? IMO, this is a functional test case and https://github.com/pingcap/automated-tests/tree/master/ticases seems more suit for writing this kind of case since TiPocket usually focuses on the stability testing on TiDB which means it runs a longer period with chaos inject randomly often.

I may have been mislead when it was suggested to use TiPocket. I'm happy to move it to automated-tests -- I had no idea this existed.

PR #415, opened 2021-06-01
cdc-bank: fix end time might be null issue.

JinLingChristopher:

fix end time might be null for create table ddl.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/tests/cdc-bank/bank.go
+++ b/tests/cdc-bank/bank.go
@@ -202,6 +202,7 @@ func (c *client) Start(ctx context.Context, cfg interface{}, clientNodes []clust
 		}(connID)
 	}
 
+	var valid, total int64
 	// downstream validators
 	for validatorIdx := 0; validatorIdx < validateConcurrency; validatorIdx++ {
 		wg.Add(1)
@@ -217,6 +218,8 @@ func (c *client) Start(ctx context.Context, cfg interface{}, clientNodes []clust
 					waitCancel()
 					log.Info("ddl synced")
 
+					atomic.AddInt64(&total, 1)
+
 					endTs, err := getDDLEndTs(downstream, tblName)
 					if err != nil {
 						log.Fatalf("[cdc-bank] get ddl end ts error %v", err)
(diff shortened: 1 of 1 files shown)


tests/cdc-bank/bank.go
```diff
@@ -345,7 +345,11 @@ func getDDLEndTs(db *sql.DB, tableName string) (result string, err error) {
 			return "", err
 		}
 		if line.JobType == "create table" && line.TblName == tableName && line.State == "synced" {
-			return line.EndTime, nil
+			if line.EndTime == nil {
+				log.Warnf("ddl end time is null, line=%+v", line)
+				return "", nil
```
zyguan, line 350: Should we continue if the endTs is empty [here](https://github.com/pingcap/tipocket/blob/c63af1c8f1f5ed86297bff9a7c0567d0565058e3/tests/cdc-bank/bank.go#L220)? Otherwise we set `@@tidb_snapshot` to an empty string, it might be unexpected.

tests/cdc-bank/bank.go
```diff
@@ -345,7 +345,11 @@ func getDDLEndTs(db *sql.DB, tableName string) (result string, err error) {
 			return "", err
 		}
 		if line.JobType == "create table" && line.TblName == tableName && line.State == "synced" {
-			return line.EndTime, nil
+			if line.EndTime == nil {
+				log.Warnf("ddl end time is null, line=%+v", line)
+				continue
```
zyguan, line 350: It's not right to continue here, if so, `getDDLEndTs` finally returns error(cannot find in ddl history, tableName ... ). I mean, the caller of `getDDLEndTs` need to handle the situation where the result is an empty string.

JinLingChristopher: if meet empty end time, just ignore that round

zyguan (approved):
LGTM

zyguan (approved):
LGTM

PR #417, opened 2021-06-04
cross-region: only fail the test if a critical error occurs

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

Currently, there are many test failures that are caused by timeouts due to resource constraints. To improve test success rate and focus on the TSO consistency check, we need to handle this situation.

- Only fail the cross-region test if a critical error occurs.
- Wait for leader election without any timeout.

### Check List 

Tests 

 - Unit test
 - E2E test

```release-note
NONE
```

### DIFF
--- a/testcase/cross-region/tso.go
+++ b/testcase/cross-region/tso.go
@@ -53,7 +53,7 @@ func (cte *criticalTSOError) Error() string {
 	case suffixChangedError:
 		errTypeMsg = "suffixChangedError"
 	}
-	return fmt.Sprintf("%s: %s", errTypeMsg, cte.msg)
+	return fmt.Sprintf("criticalTSOError: %s, %s", errTypeMsg, cte.msg)
 }
 
 // TSO wraps the tso response
--- a/testcase/cross-region/tso.go
+++ b/testcase/cross-region/tso.go
@@ -406,7 +406,7 @@ func (c *crossRegionClient) waitAllocatorReady(dcLocations []string) error {
 
 func (c *crossRegionClient) waitLeader(name string) error {
 	for {
-		fmt.Printf("Wait for %s leader election\n", name)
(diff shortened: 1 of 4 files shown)


Yisaer (commented):
I think we can set tikv/tidb replicas to 0 currently.

testcase/cross-region/tso.go
```diff
@@ -352,31 +405,33 @@ func (c *crossRegionClient) waitAllocatorReady(dcLocations []string) error {
 }
 
 func (c *crossRegionClient) waitLeader(name string) error {
-	return util2.WaitUntil(fmt.Sprintf("%s leader election", name), func() bool {
+	for {
```
Yisaer, line 408: There will be no timeout

testcase/cross-region/tso.go
```diff
@@ -352,31 +405,33 @@ func (c *crossRegionClient) waitAllocatorReady(dcLocations []string) error {
...
 		}
-		return true
-	})
+		time.Sleep(30 * time.Second)
+	}
 }
 
 func (c *crossRegionClient) waitAllocator(name, dcLocation string) error {
-	return util2.WaitUntil(fmt.Sprintf("%s of %s allocator election", name, dcLocation), func() bool {
+	for {
```
Yisaer, line 422: ditto.

Yisaer (commented):
Are the replicas of tikv/tidb set to 0?

testcase/cross-region/tso.go
JmPotato: I think it's unnecessary since the CI test of TiPocket will kill it automatically when it's timeout.

JmPotato:
@Yisaer PTAL.

Yisaer (approved):

PR #418, opened 2021-06-07
Add minimal privileges test for backup using dynamic privs

morgo:

This reopens https://github.com/pingcap/tipocket/pull/411

I discussed it with others, and there is a limitation in ticases in that it's not possible to change the DSN and effectively test the privilege system.

This depends on https://github.com/pingcap/tidb/pull/25039 merging. If Ticases can support change user, we can migrate the test at a later point.

### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

Side effects

 - None
 
Related changes

 - None

There is no imcompatibility or change - if users want to keep allocating SUPER privilege to backup operators, they can still do that. However, we should encourage principal of least privilege to our users for security best practice. So from that perspective, there is a user-facing change that using `BACKUP_ADMIN` and `RESTORE_ADMIN` will be encouraged over `SUPER`.


testcase/dynamicprivs/go.mod
```diff
@@ -0,0 +1,48 @@
+module github.com/pingcap/tipocket/testcase/dynamicprivs
+
+go 1.15
```
bb7133, line 3:
```suggestion
go 1.16
```
Please bump the version to 1.16

testcase/dynamicprivs/client.go
```diff
@@ -0,0 +1,133 @@
...
+// TearDown implements the core.Client interface.
+func (c *Client) TearDown(ctx context.Context, nodes []cluster.ClientNode, idx int) error {
+	// delete the tempdir.
+	os.RemoveAll(c.tempdir)
+	return nil
+}
+
+// Start implements the core.StandardClientExtensions interface.
+func (c *Client) Start(ctx context.Context, cfg interface{}, clientNodes []cluster.ClientNode) error {
+
```
bb7133, line 96: Please remove the blank line.

PR #432, opened 2021-09-07
do not fail if resolve lock returns region unavailable

vivid392845427:

the testcase return error message as follow，because nemesis：pd-scheduler random-merge-scheduler

2021/09/07 05:23:35 pd_scheduler.go:59: [info] apply nemesis pd-scheduler random-merge-scheduler on ns tipocket-debug-nrrdr
2021/09/07 05:23:35 pd_scheduler.go:63: [info] add scheduler random-merge-scheduler successfully
2021/09/07 05:27:34 resolve_lock.go:233: [error] [round-35] failed to run GC at safe point 427552777833021443
2021/09/07 05:27:34 resolve_lock.go:194: [info] test end 
2021/09/07 05:27:34 control.go:285: [fatal] run client error, GET request "http://debug-tidb.tipocket-debug-nrrdr.svc:10080/test/gc/resolvelock?safepoint=427552777833021443&physical=false", got 400 resolveLocks failed: region unavailable

none

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/testcase/resolve-lock/resolve_lock.go
+++ b/testcase/resolve-lock/resolve_lock.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -230,8 +231,10 @@ func (c *resolveLockClient) Start(ctx context.Context, cfg interface{}, clientNo
 		// Invoke GC with the safe point
 		greenGCUsed, err := c.resolveLocks(ctx)
 		if err != nil {
-			log.Errorf("[round-%d] failed to run GC at safe point %v", loopNum, c.safePoint)
-			return errors.Trace(err)
+			if !strings.Contains(err.Error(), "region unavailable") {
+				log.Errorf("[round-%d] failed to run GC at safe point %v", loopNum, c.safePoint)
+				return errors.Trace(err)
+			}
 		}
 		log.Infof("[round-%d] GC done at safePoint(%v)", loopNum, c.safePoint)
 
--- a/testcase/resolve-lock/resolve_lock.go
+++ b/testcase/resolve-lock/resolve_lock.go
@@ -229,12 +229,20 @@ func (c *resolveLockClient) Start(ctx context.Context, cfg interface{}, clientNo
 		}
 		log.Infof("[round-%d] start to GC at safePoint(%v)", loopNum, c.safePoint)
 		// Invoke GC with the safe point
-		greenGCUsed, err := c.resolveLocks(ctx)
-		if err != nil {
+		var greenGCUsed bool
+		for i := 0; i < 5; i++ {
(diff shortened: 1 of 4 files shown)


testcase/resolve-lock/resolve_lock.go
```diff
@@ -230,8 +231,10 @@ func (c *resolveLockClient) Start(ctx context.Context, cfg interface{}, clientNo
 		// Invoke GC with the safe point
 		greenGCUsed, err := c.resolveLocks(ctx)
 		if err != nil {
-			log.Errorf("[round-%d] failed to run GC at safe point %v", loopNum, c.safePoint)
-			return errors.Trace(err)
+			if !strings.Contains(err.Error(), "region unavailable") {
+				log.Errorf("[round-%d] failed to run GC at safe point %v", loopNum, c.safePoint)
+				return errors.Trace(err)
+			}
```
zyguan, line 237: @youjiali1995 Is it OK to check data if `resolveLocks` failed? Should we continue to the next round?

youjiali1995: > Should we continue to the next round?
Yes.

youjiali1995: I prefer to retry for some times in a loop. If all retry fails, abort the test.
```go
		var greenGCUsed bool
		for i := 0; i < 5; i++ {
			greenGCUsed, err = c.resolveLocks(ctx)
			if err == nil {
				break
			}
			if !strings.Contains(err.Error(), "region unavailable") {
				log.Errorf("[round-%d] failed to run GC at safe point %v", loopNum, c.safePoint)
				return errors.Trace(err)
			}
		}
		if err != nil {
			log.Errorf("[round-%d] failed to resolve locks for 5 times at safe point %v", loopNum, c.safePoint)
			return errors.Trace(err)
		}
```

youjiali1995 (approved):

zyguan (approved):
LGTM

testcase/resolve-lock/resolve_lock.go
```diff
@@ -540,7 +552,9 @@ func (c *resolveLockClient) reset(ctx context.Context) {
 
 func (c *resolveLockClient) getTs(ctx context.Context) (uint64, error) {
 	physical, logical, err := c.pd.GetTS(ctx)
-	if err != nil {
+	//if pd down, errmsg is io.EOF,testcase ignore this errmsg
+	//panic_check plugin check pd panic
+	if err != nil && err != io.EOF {
```
youjiali1995, line 557: It's not suitable to keep running the test if it happens.

youjiali1995: If the error is `EOF`, it means PD is down. I think we needn't handle such a case. It must be a bug.

PR #433, opened 2021-09-10
add error stack for resolve-lock test

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>

1. Add the error stack to distinguish the failure cause easily.
2. Don't keep running the test if PD is down.

### DIFF
--- a/testcase/resolve-lock/resolve_lock.go
+++ b/testcase/resolve-lock/resolve_lock.go
@@ -5,7 +5,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


zyguan (approved):
LGTM

youjiali1995:
I changed the way to handle PD's `EOF` error accoriding to @vivid392845427 words. Now if the test gets the `EOF` error, it will backoff for a while and if backoff times out, the test finishes with sucess to avoid false-negative caused by the test environment. @zyguan @vivid392845427

vivid392845427:
LGTM

zyguan (approved):

PR #435, opened 2021-09-17
fix comparing getTS error

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>

PD client wraps the raw error so that the error comparison fails. https://github.com/tikv/pd/blob/1db2398d1a508a1679ef51f5e06133cc821becbd/client/client.go#L949


vivid392845427:
LGTM

vivid392845427 (approved):

